PR #18, opened 2020-11-27
Add circular buffer example

### LINKED ISSUE #13: Prepare a simple example to save a vector timeseries in a circular buffer

Nicogene:
It fixes #13 

This is a very simple example of usage of the circular buffer
Moreover, I have added some improvements in the cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,8 +26,6 @@ if(MSVC)
   set(CMAKE_DEBUG_POSTFIX "d")
 endif()
 
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # Build position independent code.
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
@@ -53,7 +51,7 @@ endif()
 
 
 set(YARP_FORCE_DYNAMIC_PLUGINS TRUE CACHE INTERNAL "${PROJECT_NAME} is always built with dynamic plugins")
-find_package(YARP 3.4 COMPONENTS os conf REQUIRED)
+find_package(YARP 3.4 COMPONENTS conf REQUIRED)
 # Use the same build options as YARP
 set(BUILD_SHARED_LIBS ${YARP_IS_SHARED_LIBRARY})
 set(CMAKE_C_FLAGS ${YARP_C_FLAGS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,8 @@ if(MSVC)
   set(CMAKE_DEBUG_POSTFIX "d")
 endif()
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # Build position independent code.
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,9 @@
 # BSD-3-Clause license. See the accompanying LICENSE file for details.
 
 cmake_minimum_required(VERSION 3.12)
-project(yarp-telemetry)
+project(yarp-telemetry
+       VERSION 0.0.1
+       LANGUAGES CXX)
 
 include(GNUInstallDirs)
 include(FeatureSummary)
@@ -49,7 +51,7 @@ endif()
 
 
 set(YARP_FORCE_DYNAMIC_PLUGINS TRUE CACHE INTERNAL "${PROJECT_NAME} is always built with dynamic plugins")
-find_package(YARP 3.4 COMPONENTS conf REQUIRED)
+find_package(YARP 3.4 COMPONENTS os conf REQUIRED)
 # Use the same build options as YARP
 set(BUILD_SHARED_LIBS ${YARP_IS_SHARED_LIBRARY})
 set(CMAKE_C_FLAGS ${YARP_C_FLAGS})
@@ -58,6 +60,7 @@ set(CMAKE_CXX_FLAGS ${YARP_CXX_FLAGS})
 
 #### INSERT find_package(MATIO) and other libraries here
 find_package(matioCpp REQUIRED)
+find_package(Boost REQUIRED COMPONENTS filesystem system)
 
 
 feature_summary(WHAT ALL INCLUDE_QUIET_PACKAGES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ set(CMAKE_CXX_FLAGS ${YARP_CXX_FLAGS})
 
 #### INSERT find_package(MATIO) and other libraries here
 find_package(matioCpp REQUIRED)
-find_package(Boost REQUIRED COMPONENTS filesystem system)
+find_package(Boost REQUIRED)
 
 
 feature_summary(WHAT ALL INCLUDE_QUIET_PACKAGES)
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -6,4 +6,4 @@
 
 set(CIRCULAR_BUFFER_EXAMPLE_SRC circular-buffer-example.cpp)
 add_executable(circular-buffer-example ${CIRCULAR_BUFFER_EXAMPLE_SRC})
-target_link_libraries(circular-buffer-example Boost::system)
+target_link_libraries(circular-buffer-example Boost::boost)
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -4,5 +4,6 @@
 # This software may be modified and distributed under the terms of the
 # BSD-3-Clause license. See the accompanying LICENSE file for details.
 
-add_subdirectory(libYARP_telemetry)
-add_subdirectory(examples)
+set(CIRCULAR_BUFFER_EXAMPLE_SRC circular-buffer-example.cpp)
+add_executable(circular-buffer-example ${CIRCULAR_BUFFER_EXAMPLE_SRC})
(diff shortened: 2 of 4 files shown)


src/examples/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Copyright (C) 2006-2020 Istituto Italiano di Tecnologia (IIT)
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+set(CIRCULAR_BUFFER_EXAMPLE_SRC circular-buffer-example.cpp)
+add_executable(circular-buffer-example ${CIRCULAR_BUFFER_EXAMPLE_SRC})
+target_link_libraries(circular-buffer-example Boost::system)
```
traversaro, line 9: If you are using a header-only boost library, you can use the `Boost::headers` imported target.

CMakeLists.txt
```diff
@@ -58,6 +62,7 @@ set(CMAKE_CXX_FLAGS ${YARP_CXX_FLAGS})
 
 #### INSERT find_package(MATIO) and other libraries here
 find_package(matioCpp REQUIRED)
+find_package(Boost REQUIRED COMPONENTS filesystem system)
```
traversaro, line 65: If you only use a header-only library, I think we can just do:
```suggestion
find_package(Boost REQUIRED)
```

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,51 @@
...
+    cb.pop_front(); // 3 is removed.
+
+    cout<<"Removing 3 and 5"<<std::endl;
+    cout<<"The circular buffer contains:"<<endl;
+    for (auto& c_el : cb) {
+        cout<<c_el<<endl;
+    }
+
+    return 0;
+ }
```
traversaro, line 51: One thing that I don't get of this example, is that the main advantage of a circular buffer is that you can reuse an existing data structure without allocating or de-allocating it, but I think in this specific implementation we don't have this property, right?

src/examples/CMakeLists.txt
Nicogene: I think it is `Boost::boost` the target containing the headers if I remember correctly

src/examples/circular-buffer-example.cpp
Nicogene: In this case we have a buffer of 3 elements and we call `push_back` 5 times and here it is the output:

```
The circular buffer contains:
1606490129 1
1606490130 2
1606490131 3
The circular buffer contains:
1606490131 3
1606490132 4
1606490133 5
Removing 3 and 5
The circular buffer contains:
1606490132 4
```

No re-allocation has been done, when pushing 4 and 5 the oldest data(1 and 2) has been replaced

traversaro: Ah ok. However I was confused because I was thinking of the case in which you first `pop` an element, and then you add another one, but that is probably not how the circular buffer is typically used.

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }  
+
+};
+
+ int main()
+ {
+    // Create a circular buffer with a capacity for 3 integers.
+    boost::circular_buffer<Data> cb(3);
+
+    // Insert threee elements into the buffer.
```
drdanz, line 31: 3 with 3 'e' is very representative of the number 3! :laughing:

src/examples/circular-buffer-example.cpp
drdanz: I think it would be good to add to `struct Data` a print in the constructors (the one you added, copy, and move), in copy and move operators and in destructor, in order to evaluate the number of copies/moves performed by the circular buffer

CMakeLists.txt
```diff
@@ -24,6 +26,8 @@ if(MSVC)
   set(CMAKE_DEBUG_POSTFIX "d")
 endif()
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
drdanz, line 30: This is "deprecated"[*], the recommended way to do this is to add
```cmake
target_compile_features(foo [PUBLIC|PRIVATE|INTERFACE] cxx_std_14)
```
to the target `foo`, which will export the requirements to the the targets that depend on this target.


[*] (I'm not sure if this is actually deprecated, but it is not the modern cmake way to do it)

CMakeLists.txt
```diff
@@ -49,7 +53,7 @@ endif()
 
 
 set(YARP_FORCE_DYNAMIC_PLUGINS TRUE CACHE INTERNAL "${PROJECT_NAME} is always built with dynamic plugins")
-find_package(YARP 3.4 COMPONENTS conf REQUIRED)
+find_package(YARP 3.4 COMPONENTS os conf REQUIRED)
```
drdanz, line 56: This is probably not needed, at least for now, until we actually want to add a dependency on `os`

src/examples/circular-buffer-example.cpp
drdanz: Which suddenly reminded me of [Gobliiins](https://en.wikipedia.org/wiki/Gobliiins)!

traversaro:
@AlexAntn did you had the occasion to check out the example prepare by @Nicogene ?

CMakeLists.txt
Nicogene: Yes it was a leftover because, in the beginning, I used `yarp::os::Time::now()` but I thought to avoid using YARP just for the clock

CMakeLists.txt
Nicogene: Thanks. I didn't know it

src/examples/circular-buffer-example.cpp
Nicogene: Good point

Nicogene:
I added the prints in the constructors/operators as suggested and here is the output

```
Data(_ts, _datum)
Data(Data&&)
~Data
Data(_ts, _datum)
Data(Data&&)
~Data
Data(_ts, _datum)
Data(Data&&)
~Data
The circular buffer contains:
1606903056 1
1606903057 2
1606903058 3
Data(_ts, _datum)
operator=(Data&&)
~Data
Data(_ts, _datum)
operator=(Data&&)
~Data
The circular buffer contains:
1606903058 3
1606903059 4
1606903060 5
~Data
~Data
Removing 3 and 5
The circular buffer contains:
1606903059 4
~Data
```

src/examples/circular-buffer-example.cpp
Nicogene: I will leave it as it is as an easter egg 😆

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,96 @@
...
+    cb.push_back(Data(std::chrono::system_clock::now(), 1));
```
drdanz, line 63: This line causes
```
Data(_ts, _datum)
Data(Data&&)
~Data
```

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,96 @@
...
+    cb.push_back(Data(std::chrono::system_clock::now(), 3));
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+
+    cout<<"The circular buffer contains:"<<endl;
+    for (auto& c_el : cb) {
+        cout<<std::chrono::system_clock::to_time_t(c_el.ts)<<" "<<c_el.datum<<endl;
+    }
+
+
+    cb.push_back(Data(std::chrono::system_clock::now(), 4));
```
drdanz, line 76: This line causes:
```
Data(_ts, _datum)
operator=(Data&&)
~Data
```

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,96 @@
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+    cb.push_back(Data(std::chrono::system_clock::now(), 5));
+
+
+    cout<<"The circular buffer contains:"<<endl;
+    for (auto& c_el : cb) {
+        cout<<std::chrono::system_clock::to_time_t(c_el.ts)<<" "<<c_el.datum<<endl;
+    }
+    // Elements can be popped from either the front or the back.
+    cb.pop_back();  // 5 is removed.
```
drdanz, line 86: This line causes
```
~Data
```

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,96 @@
...
+    cb.pop_front(); // 3 is removed.
+
+    cout<<"Removing 3 and 5"<<std::endl;
+    cout<<"The circular buffer contains:"<<endl;
+    for (auto& c_el : cb) {
+        cout<<std::chrono::system_clock::to_time_t(c_el.ts)<<" "<<c_el.datum<<endl;
+    }
+
+    return 0;
+ }
```
drdanz, line 96: Finally the last `Data` is deleted
```
~Data
```

drdanz (commented):
Not a review, just commenting on each line to see what is happening.

src/examples/circular-buffer-example.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    ~Data(){
+        std::cout<<"~Data"<<std::endl;
+    }
+
+};
+
+ int main()
+ {
+    // Create a circular buffer with a capacity for 3 integers.
```
AlexAntn, line 59: I'm guessing after introducing the struct `Data`, the buffer now has the capacity for 3 entries of type `Data` and not integers, correct?

Nicogene: Yes! It is a leftover of the first implementation

traversaro (approved):

PR #22, opened 2020-12-07
Add record api

### LINKED ISSUE #19: Prepare data structure for element of the circular buffer

Nicogene:
It is the very first draft of what the telemetry "record" could be.

I also added an example that exploits that API.

It fixes #19 

cc @S-Dafarra


src/examples/circular_buffer_record_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+using namespace std;
+using namespace yarp::os;
+using namespace yarp::telemetry;
+
+ int main()
+ {
+    Network yarp;
+
+    std::cout<<"XXXXXXXX CIRCULAR BUFFER OF INT XXXXXXXX"<<std::endl;
+    // Create a circular buffer with a capacity for 3 Data structures.
```
AlexAntn, line 30: the capacity here is for 3 Ints, not Data structures 😄

src/examples/circular_buffer_record_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+    for (auto& c_el : cb_i) {
+        total_payload += c_el.getPayload();
+        cout<<c_el.m_ts.getCount() << " " << std::setw( 14 ) << std::setprecision( 14 ) << c_el.m_ts.getTime() << " | " << c_el.m_datum<<std::endl;
+    }
+
+    std::cout<<"The circular buffer payload is "<<total_payload<<" bytes "<<std::endl;
+
+    std::cout<<"XXXXXXXX CIRCULAR BUFFER OF DOUBLE XXXXXXXX"<<std::endl;
+
+    // Create a circular buffer with a capacity for 3 Data structures.
```
AlexAntn, line 67: Same as the comment above, but doubles

src/examples/circular_buffer_record_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+    cout<<"The circular buffer contains:"<<endl;
+    for (auto& c_el : cb_i) {
+        total_payload += c_el.getPayload();
+        cout<<c_el.m_ts.getCount() << " " << std::setw( 14 ) << std::setprecision( 14 ) << c_el.m_ts.getTime() << " | " << c_el.m_datum<<std::endl;
+    }
+
+    std::cout<<"The circular buffer payload is "<<total_payload<<" bytes "<<std::endl;
+    std::cout<<"XXXXXXXX CIRCULAR BUFFER OF VECTOR OF DOUBLE XXXXXXXX"<<std::endl;
+
+    // Create a circular buffer with a capacity for 3 Data structures.
```
AlexAntn, line 103: Same as the comment above, but vectors of doubles

src/examples/circular_buffer_record_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+        {
+             cout<<v_el<<" ";
+        }
+        cout<<std::endl;
+    }
+
+    std::cout<<"The circular buffer payload is "<<total_payload<<" bytes "<<std::endl;
+
+    return 0;
+ }
```
AlexAntn, line 149: It would be nice to have an example actually using the data structures, as in the circular buffer example!

Nicogene: Actually, they are not `int32_t`, `double` and `vector<double>` but `Record<int32_t>`, `Record<double>` and `Record<vector<double>>`

`Record` is very similar to the `Data` structure of the old example, except that uses ``yarp::os::Stamp` for the timestamp(we can discuss if it is the case to keep as it is or change to something standard) and it is a template class

Nicogene:
I know that this PR miss the documentation, but I would add it once we agreed on the API.

Here is the output of the `circular_buffer_record_example` executable:
```
XXXXXXXX CIRCULAR BUFFER OF INT XXXXXXXX
The capacity is: 3 and the size is: 0
The capacity is: 3 and the size is: 3
The circular buffer contains:
0 1607682920.9354 | 1
1 1607682921.1363 | 2
2 1607682921.3365 | 3
The capacity is: 3 and the size is: 3
The circular buffer contains:
2 1607682921.3365 | 3
3 1607682921.5367 | 4
4 1607682921.7368 | 5
The circular buffer payload is 108 bytes
XXXXXXXX CIRCULAR BUFFER OF DOUBLE XXXXXXXX
The capacity is: 3 and the size is: 0
The capacity is: 3 and the size is: 3
The circular buffer contains:
2 1607682921.3365 | 3
3 1607682921.5367 | 4
4 1607682921.7368 | 5
The capacity is: 3 and the size is: 3
The circular buffer contains:
2 1607682921.3365 | 3
3 1607682921.5367 | 4
4 1607682921.7368 | 5
The circular buffer payload is 108 bytes
XXXXXXXX CIRCULAR BUFFER OF VECTOR OF DOUBLE XXXXXXXX
The capacity is: 3 and the size is: 0
The capacity is: 3 and the size is: 3
The circular buffer contains:
0 1607682922.9384 | 0.1 0.2 0.3
1 1607682923.1389 | 0.3 0.4 0.5
2 1607682923.3394 | 0.6 0.7 0.8
The capacity is: 3 and the size is: 3
The circular buffer contains:
2 1607682923.3394 | 0.6 0.7 0.8
3 1607682923.5396 | 0.9 1 1.1
4 1607682923.7398 | 1.2 1.3 1.4
The circular buffer payload is 168 bytes
```

src/examples/circular_buffer_record_example.cpp
AlexAntn: Ah, that is correct! 

Regarding the time stamp, it would be nicer if we could detach it from yarp (but allow, for example, to initialize with yarp stamp or something else), but we can discuss if it makes sense.

traversaro (approved):
Thanks @Nicogene ! I think it make sense. Just a doubt, what is the role of the `m_payload`  attribute?

AlexAntn (approved):

Nicogene:
>  I think it make sense. Just a doubt, what is the role of the m_payload attribute?

I initially added it as debug variable, but it may become useful when deciding the length of the buffer, but if not needed I will remove it!

src/libYARP_telemetry/src/yarp/telemetry/Record.h
```diff
@@ -0,0 +1,47 @@
...
+template<class T>
+struct Record
+{
+    yarp::os::Stamp m_ts;
+    T m_datum;
+
+
+    Record(const yarp::os::Stamp& _ts,
+           const T& _datum) : m_ts(_ts), m_datum(_datum) {
+               m_payload = sizeof(m_ts) + sizeof(m_datum);
```
S-Dafarra, line 25: If this working with containers? I guess that in case of vectors this does not compute the total payload :thinking:

Nicogene: Yes you are right, for example see this code snippet:

http://cpp.sh/5utpi

The `sizeof(vector<double>)` is always 24, independently by the number of elements

Nicogene:
I opened an issue for adressing the payload problem(https://github.com/robotology-playground/yarp-telemetry/issues/23) merging as it is, that issue will be addressed later

PR #27, opened 2021-01-07
Pipeline example

AlexAntn:
After working with Nicolo in the morning, we came up with a small example of the full pipeline, from a data vector to circular buffer, then reading from the circular buffer and creating a mat file with matioCpp.

The reading is done in a thread, made safe by use of mutex locks. The data is stored in a local collection, and the mat file is created at the end of the run (when the user chooses to terminate the module).

This already uses the Buffer API developed by @Nicogene

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ set(CMAKE_CXX_FLAGS ${YARP_CXX_FLAGS})
 #### INSERT find_package(MATIO) and other libraries here
 find_package(matioCpp REQUIRED)
 find_package(Boost REQUIRED)
+find_package(Threads REQUIRED)
 
 
 feature_summary(WHAT ALL INCLUDE_QUIET_PACKAGES)
--- a/src/examples/CB_to_matfile_example.cpp
+++ b/src/examples/CB_to_matfile_example.cpp
@@ -1,0 +1,230 @@
+#include <boost/circular_buffer.hpp>
+#include <iostream>
+#include <chrono>
+#include <thread>
+#include <random>
+#include <matioCpp/matioCpp.h>
+#include <mutex>
+
+#include <yarp/telemetry/Buffer.h>
+#include <yarp/os/Time.h>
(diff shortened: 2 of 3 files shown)


traversaro (approved):

traversaro:
The homebrew failure is not related to this repo, but rather to https://github.com/actions/virtual-environments/issues/2391 or https://github.com/actions/virtual-environments/issues/2322 . We can add some workarounds, but given the frequency of this incidents I think we can also just disable macOS testing for now.

traversaro:
@AlexAntn can you please write down the discussion in today's meeting? Thanks!

AlexAntn:
In the meeting, we discussed the use of a circular buffer, and whether it makes sense to copy the contents read from a buffer to a "local collection" vector. 

In practice, if we make the buffer large enough and also specify a threshold at which it writes the data to a file, it should be possible to work just with the circular buffer (for example, the circular buffer could have a capacity for 300 entries of data, and a threshold for saving to the file at 200, allowing still 100 entries to write new data without losing any data).

Another aspect is the saving of the file when something happens (e.g.: ctrl+c or a segmentation fault exception). This might require using yarp modules or checking how to do it in a OS-independent way.

I will implement these changes and update the PR.

Nicogene:
As stated [here](https://github.com/robotology-playground/yarp-telemetry/pull/31#issue-554070040) the ideas that came out from the latest meeting has been implemented in the high-level API.
In this sense, I would merge this PR as it is since this example already accomplished its purpose.

cc @AlexAntn @traversaro @S-Dafarra @drdanz

traversaro:
> 
> 
> As stated [here](https://github.com/robotology-playground/yarp-telemetry/pull/31#issue-554070040) the ideas that came out from the latest meeting has been implemented in the high-level API.
> In this sense, I would merge this PR as it is since this example already accomplished its purpose.
> 
> cc @AlexAntn @traversaro @S-Dafarra @drdanz

Ok for me!

PR #31, opened 2021-01-13
Add BufferManager API

### LINKED ISSUE #23: Fix/investigate a way to express the payload of a single Record
In #22 it has been added just as a sum of the sizeof of the members of `Record` but this does not work with the containers(e.g. see http://cpp.sh/5utpi)

This kind of info may be useful for defining a RAM cap occupied by the circular buffer and then its maximum length in function of which kind of data is carrying

### LINKED ISSUE #29: Define and implement high-level user API
After we understood the basic building blocks, we should define a concrete API that our users can use to easily specify telemetry log for signals.

Inspiration could be given by: 
* https://github.com/robotology-playground/yarp-telemetry/issues/17 

From https://github.com/robotology-playground/yarp-telemetry/pull/27 we learned: 
* Specify if a periodic thread to dump data to file is desired, or if no thread is used and an external function is called to create the file (for example at the end of the experiment)

Nicogene:
It fixes #29 
It fixes #23
 
This PR introduces these changes:

- Use `double` as timestamp in `Record<T>` (see https://github.com/robotology-playground/yarp-telemetry/issues/28#issuecomment-758003434)
- Use `std::vector<T>` as type for `m_datum` of `Record<T>`(see https://github.com/robotology-playground/yarp-telemetry/issues/23#issuecomment-756782927).
- Implements https://github.com/robotology-playground/yarp-telemetry/pull/27#issuecomment-758001061
- Add `BufferManager` class

This is a template class that contains a `std::map<string, telemetry::Buffer<T>>` that contains all the user-defined `Buffer`s, each one associated to a variable.
Each `BufferManager` defines a .mat structure containing all the variables.

For now it DOES NOT handle the multithreading.

Here is the example of scalar signals:

```c++
yarp::telemetry::BufferManager<int32_t> bm({ {"one",{1,1}},
                                                 {"two",{1,1}} }, 3);

for (int i = 0; i < 10; i++) {
    bm.push_back({ i }, "one");
    yarp::os::Time::delay(0.2);
    bm.push_back({ i + 1 }, "two");
}
```

That produces:

```
buffer_manager_test = 

  struct with fields:

    one: [1×1 struct]
    two: [1×1 struct]


buffer_manager_test.one

ans = 

  struct with fields:

          data: [1×1×3 int32]
    dimensions: [3 1 1]
          name: 'one'
    timestamps: [523132.9969457 523133.1979436 523133.3988861]
```
Here is the example of vector signals

```c++
yarp::telemetry::BufferManager<double> bm_v({ {"one",{4,1}},
                                              {"two",{4,1}} }, 3);

for (int i = 0; i < 10; i++) {
    bm_v.push_back({ i+1.0, i+2.0, i+3.0, i+4.0  }, "one");
    yarp::os::Time::delay(0.2);
    bm_v.push_back({ (double)i, i*2.0, i*3.0, i*4.0 }, "two");
}



```
That produces

```
buffer_manager_test_vector = 

  struct with fields:

    one: [1×1 struct]
    two: [1×1 struct]

>> buffer_manager_test_vector.one

ans = 

  struct with fields:

          data: [4×1×3 double]
    dimensions: [3 4 1]
          name: 'one'
    timestamps: [523135.0186688 523135.219639 523135.4203739]
```

Please review code.

cc @S-Dafarra

### DIFF
--- a/src/examples/telemetry_buffer_manager_example.cpp
+++ b/src/examples/telemetry_buffer_manager_example.cpp
@@ -48,7 +48,6 @@ using namespace yarp::telemetry;
         bm_v.push_back({ (double)i, i*2.0, i*3.0, i*4.0 }, "two");
     }
 
-    // ISSUES HERE
     if (bm_v.saveToFile("buffer_manager_test_vector.mat"))
         std::cout << "File saved correctly!" << std::endl;
     else
--- a/src/examples/telemetry_buffer_manager_example.cpp
+++ b/src/examples/telemetry_buffer_manager_example.cpp
@@ -39,7 +39,20 @@ using namespace yarp::telemetry;
     else
         std::cout << "Something went wrong..." << std::endl;
 
+    yarp::telemetry::BufferManager<vector<double>> bm_v({ {"one",{4,1}},
+                                                        {"two",{4,1}} }, 3);
 
+    for (int i = 0; i < 10; i++) {
+        bm_v.push_back({ i+1.0, i+2.0, i+3.0, i+4.0  }, "one");
+        yarp::os::Time::delay(0.2);
+        bm_v.push_back({ (double)i, i*2.0, i*3.0, i*4.0 }, "two");
+    }
+
+    // ISSUES HERE
+    //if (bm_v.saveToFile("buffer_manager_test_vector.mat"))
+    //    std::cout << "File saved correctly!" << std::endl;
+    //else
+    //    std::cout << "Something went wrong..." << std::endl;
 
     return 0;
  }
--- a/src/examples/telemetry_buffer_manager_example.cpp
+++ b/src/examples/telemetry_buffer_manager_example.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2006-2020 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+
+
+#include <yarp/os/Time.h>
+#include <yarp/os/Network.h>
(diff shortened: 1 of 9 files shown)


src/examples/telemetry_buffer_manager_example.cpp
```diff
@@ -0,0 +1,58 @@
...
+    yarp::telemetry::BufferManager<double> bm_v({ {"one",{4,1}},
+                                                  {"two",{4,1}} }, 3);
+
+    for (int i = 0; i < 10; i++) {
+        bm_v.push_back({ i+1.0, i+2.0, i+3.0, i+4.0  }, "one");
+        yarp::os::Time::delay(0.2);
+        bm_v.push_back({ (double)i, i*2.0, i*3.0, i*4.0 }, "two");
+    }
+
+    // ISSUES HERE
```
traversaro, line 51: Leftover?

src/libYARP_telemetry/src/CMakeLists.txt
```diff
@@ -52,7 +53,7 @@ target_compile_features(YARP_telemetry PUBLIC cxx_std_17)
 target_link_libraries(YARP_telemetry PUBLIC YARP::YARP_conf
                                             YARP::YARP_os
                                             Boost::boost
-                                     PRIVATE matioCpp::matioCpp)
+                                     matioCpp::matioCpp) # This has to become PRIVATE
```
traversaro, line 56: If you want to track some requirement or desiderata, it probably it is easier to open an issue so we can track it from the project management perspective.

traversaro (approved):
Minor comments. I think it is a step in the right direction! For me it is ok to merge and try to use it somewhere (https://github.com/robotology-playground/yarp-telemetry/issues/26), and if improvements are possible we will see them there.

PR #42, opened 2021-01-22
BufferManager: add auto save option

### LINKED ISSUE #35: Add an option to save the file in the destructor of the BufferManager

Nicogene:
As we discussed in the last meeting we decided to implement the auto-save option in order to save the `.mat` file in the destructor.

The option is **false by default** but this can be changed.
Moreover, I also changed the example that uses the `BufferManager` in order to exploit this autosave option.

It fixes #35 

Please review code.

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -36,10 +36,10 @@ public:
     BufferManager() = delete;
     BufferManager(const std::string& filename, const std::vector<BufferInfo>& listOfVars,
                   size_t n_samples, bool auto_save=false) : m_filename(filename), m_auto_save(auto_save) {
-	    assert(listOfVars.size() != 0);
+        assert(listOfVars.size() != 0);
         assert(!filename.empty());
-		for (const auto& s : listOfVars) {
-			m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(s.m_var_name,Buffer<T>(n_samples)));
+        for (const auto& s : listOfVars) {
+            m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(s.m_var_name,Buffer<T>(n_samples)));
             m_dimensions_map.insert(std::pair<std::string, yarp::telemetry::dimensions_t>(s.m_var_name, s.m_dimensions));
 		}
 	}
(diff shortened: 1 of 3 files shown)


traversaro (approved):
Seems nice, even if probably @S-Dafarra may look into that. However, the amount of stuff that we are putting in the constructor seems to hint that we may want to move all this configuration stuff to proper methods, such as `setDefaultFileName`, `addChannel`, etc etc.

S-Dafarra:
> Seems nice, even if probably @S-Dafarra may look into that. However, the amount of stuff that we are putting in the constructor seems to hint that we may want to move all this configuration stuff to proper methods, such as `setDefaultFileName`, `addChannel`, etc etc.

That's a very good point. Most of the times, you know what variables to store after the configuration phase, while the BufferManager could be a parameter of some class. Hence, in this case, it may be helpful to have a default constructor, and an "initialization" method. An alternative to the initialization method would be the move assignment, resulting in a code like:
```c++
class Foo
{
public:
    BufferManager manager;
}

//.............. In some body

Foo foo; //Here the default constructor would be needed

foo.manager = BufferManager(/* initialization */);
```

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -34,14 +34,22 @@ class BufferManager {
...
+        for (const auto& s : listOfVars) {
+            m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(s.m_var_name,Buffer<T>(n_samples)));
             m_dimensions_map.insert(std::pair<std::string, yarp::telemetry::dimensions_t>(s.m_var_name, s.m_dimensions));
 		}
 	}
 
+    ~BufferManager() {
+        if (m_auto_save) {
+            saveToFile();
+        }
```
AlexAntn, line 50: I gather the auto_save flag is only to trigger the saving when the modules are closed? I would maybe rename it to something like save_before_close or something similar - my first impression was that this would save the data everytime the buffer was full.

Anyway, minor comment

traversaro: Good point!

Nicogene: My idea was to use the same bool flag to enable the periodic save once we will implement it.
If set to true the `BufferManager` will take care of the save to file, both for when the buffer is full both when we close the application
But we can discuss it for sure

traversaro:
@Nicogene as discussed in the meeting, can you merge?

PR #52, opened 2021-01-26
BufferManager: fix order dimensions field

### LINKED ISSUE #45: Investigate strange ordering in dimensions of the saved variables

Nicogene:
After an analysis on who `matio-cpp` works with @AlexAntn we discovered that the first two dimensions have to represent the single sample, and the one regarding the number of samples is to be the last one in the list.

It fixes #45 

Moreover, we added an example using matrix variables(2x3 and 3x2)

cc @S-Dafarra

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ ans =
   struct with fields:
 
           data: [1×1×3 int32]
-    dimensions: [3 1 1]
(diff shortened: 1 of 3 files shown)


traversaro (approved):
cc @S-Dafarra

PR #53, opened 2021-01-27
BufferManager: add configuration methods and examples enhancement

Nicogene:
This PR adds some configuration methods, now you are no more forced to configure the BufferManager at the construction phase.
The examples have been cleaned up and refactored accordingly.

Please review code.

cc @S-Dafarra

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,19 +58,28 @@ target_link_libraries(myApp YARP::YARP_telemetry)
 Here is the code snippet for dumping in a `.mat` file 3 samples of the scalar varibles `"one"` and `"two"`.
 
 ```c++
-yarp::telemetry::BufferManager<int32_t> bm({ {"one",{1,1}},
-                                                 {"two",{1,1}} }, 3);
+    yarp::telemetry::BufferManager<int32_t> bm(n_samples);
+    bm.setFileName("buffer_manager_test.mat");
+    ChannelInfo var_one{ "one", {1,1} };
+    ChannelInfo var_two{ "two", {1,1} };
 
-for (int i = 0; i < 10; i++) {
-    bm.push_back({ i }, "one");
-    yarp::os::Time::delay(0.2);
-    bm.push_back({ i + 1 }, "two");
-}
+    auto ok = bm.addChannel(var_one);
+    ok = ok && bm.addChannel(var_two);
+    if (!ok) {
+        std::cout << "Problem adding variables...."<<std::endl;
+        return 1;
+    }
 
-if (bm.saveToFile("buffer_manager_test.mat"))
-    std::cout << "File saved correctly!" << std::endl;
-else
-    std::cout << "Something went wrong..." << std::endl;
+    for (int i = 0; i < 10; i++) {
+        bm.push_back({ i }, "one");
+        yarp::os::Time::delay(0.2);
+        bm.push_back({ i + 1 }, "two");
+    }
+
+    if (bm.saveToFile())
+        std::cout << "File saved correctly!" << std::endl;
+    else
+        std::cout << "Something went wrong..." << std::endl;
 
 ```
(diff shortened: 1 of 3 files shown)


src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -135,7 +161,8 @@ class BufferManager {
     }
 private:
     std::string m_filename;
-    bool m_auto_save;
+    bool m_auto_save{false};
+    size_t m_n_samples{0};
     std::map<std::string, Buffer<T>> m_buffer_map;
```
S-Dafarra, line 166: I would suggest using an [``unordered_map``](https://en.cppreference.com/w/cpp/container/unordered_map). The complexity of finding an element in an ``unordered_map`` is constant, while it is logarithmic with ``map``. If you don't care about the ordering, using an ``unordered_map`` is more efficient.

README.md
```diff
@@ -58,19 +58,28 @@ target_link_libraries(myApp YARP::YARP_telemetry)
 Here is the code snippet for dumping in a `.mat` file 3 samples of the scalar varibles `"one"` and `"two"`.
 
 ```c++
-yarp::telemetry::BufferManager<int32_t> bm({ {"one",{1,1}},
-                                                 {"two",{1,1}} }, 3);
+    yarp::telemetry::BufferManager<int32_t> bm(n_samples);
+    bm.setFileName("buffer_manager_test.mat");
+    ChannelInfo var_one{ "one", {1,1} };
```
traversaro, line 63: I think `yarp::telemetry` is missing in several of this examples.

Nicogene: For now I put `yarp::telemetry` in the latest example that is the one that is in the README, we should remove all the `using namespace yarp::telemetry`

traversaro (approved):

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -135,9 +161,10 @@ class BufferManager {
     }
 private:
     std::string m_filename;
-    bool m_auto_save;
-    std::map<std::string, Buffer<T>> m_buffer_map;
-    std::map<std::string, dimensions_t> m_dimensions_map;
+    bool m_auto_save{false};
+    size_t m_n_samples{0};
+    std::unordered_map<std::string, Buffer<T>> m_buffer_map;
```
S-Dafarra, line 166: I guess that also the corresponding include should change

Nicogene: Thanks for the tip, I didn't know that there was a specific header!

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -34,14 +34,13 @@ class BufferManager {
...
+    }
+    BufferManager(const std::string& filename, const std::vector<ChannelInfo>& channels,
+                  size_t n_samples, bool auto_save=false) : m_filename(filename), m_auto_save(auto_save), m_n_samples(n_samples){
+        assert(!channels.empty());
         assert(!filename.empty());
-        for (const auto& s : listOfVars) {
-            m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(s.m_var_name,Buffer<T>(n_samples)));
-            m_dimensions_map.insert(std::pair<std::string, yarp::telemetry::dimensions_t>(s.m_var_name, s.m_dimensions));
-		}
+        assert(addChannels(channels) == true);
```
Nicogene, line 43: With @AlexAntn we spotted that is a bug, in `Release` the `addChannels` is not invoked because `assert` is skipped

PR #55, opened 2021-01-29
BufferManager: added a periodic saving method and updated examples

### LINKED ISSUE #47: Investigate automatic saving of data
See comment from https://github.com/robotology-playground/yarp-telemetry/issues/37 :

> A virtually unlimited buffer dimension. An experiment may last from 30sec to a couple of hours (as it happens during fairs or stress tests). If we are not sure this tool may back us for the entirety of the experiment, we will most likely use other tools.

### LINKED ISSUE #59: telemetry_buffer_manager_example segfaults if compiled in release
When I run `./bin/telemetry_buffer_manager_example` I got the following segmentation fault 
```
File saved correctly!
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
Aborted (core dumped)
```
```
gdb$ where
#0  __GI_raise (sig=sig@entry=0x6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6c51921 in __GI_abort () at abort.c:79
#2  0x00007ffff7a99257 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7aa4606 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7aa4671 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7aa4905 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7a9bb46 in std::__throw_out_of_range(char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x0000555555558466 in std::__detail::_Map_base<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, yarp::telemetry::Buffer<double> >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, yarp::telemetry::Buffer<double> > >, std::__detail::_Select1st, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true>, true>::at (__k="one", this=0x7fffffffc730) at /usr/include/c++/7/bits/hashtable_policy.h:774
#8  std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, yarp::telemetry::Buffer<double>, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, yarp::telemetry::Buffer<double> > > >::at (__k="one", this=0x7fffffffc730) at /usr/include/c++/7/bits/unordered_map.h:990
#9  yarp::telemetry::BufferManager<double>::push_back (var_name="one", elem=..., this=0x7fffffffc700) at /home/gromualdi/robot-code/yarp-telemetry/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h:85
#10 main () at /home/gromualdi/robot-code/yarp-telemetry/src/examples/telemetry_buffer_manager_example.cpp:75
```

I think it's related to #57

AlexAntn:
This PR introduces a periodic saving method to the BufferManager API. 

It runs in a thread that is activated periodically, checking the buffer to see if the amount of data currently stored crosses a specified threshold. If the threshold is crossed, the data is saved to a file, and the buffer is cleared and ready to accept more data.

This is a first step towards ensuring no data loss since we are now not bound by the size of the buffer. The file names are incremented at each save (so, e.g.: `buffer_test_0.mat`, `buffer_test_1.mat`, etc). 

The saving of the file is triggered if any channel crosses the data amount threshold. This ensures that the same .mat file corresponds to the same time-window, even if the number of entries between the channels doesn't match (e.g.: if the channels have different frequencies). 

One issue we found with the current implementation is the case where the save triggers just before a data push - this can lead to the loss of one entry, with the current buffer clearing. In this example we lost an entry from channel "two":

Screenshot from 2021-01-29 16-35-01

This could potentially be solved with a different buffer cleaning technique (instead of a full clear), or with the introduction of mutexes. 

Also, given the number of parameters being passed to the constructor, I and @Nicogene  have created a `BufferConfig` structure to encode these parameters, making it easier to initialize the `BufferManager`.

It fixes #47 
It fixes #59

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -18,6 +18,9 @@
 #include <functional>
 #include <chrono>
 #include <matioCpp/matioCpp.h>
+#include <thread>
+
+#include <chrono>
 
 
 namespace yarp::telemetry {
@@ -29,6 +32,13 @@ struct ChannelInfo {
     dimensions_t m_dimensions{ 1,1 };
 };
 
+struct BufferConfig {
+    size_t n_samples = 0;
+    unsigned int check_period = 10;
+    size_t threshold = 0;
+    bool m_auto_save = false;
+    bool save_periodically = false;
+};
 
 
 template<class T>
@@ -36,18 +46,69 @@ class BufferManager {
 
 public:
     BufferManager() = delete;
-    explicit BufferManager(size_t n_samples, bool auto_save = false) : m_n_samples(n_samples), m_auto_save(auto_save) {
+    //explicit BufferManager(size_t n_samples, unsigned int _check_period = 10, size_t _threshold = 0, bool auto_save = false, bool _save_periodically = false) : m_n_samples(n_samples), m_auto_save(auto_save), check_period(_check_period), m_threshold(_threshold), save_periodically(_save_periodically) {
+    explicit BufferManager(const BufferConfig& _bufferConfig) : bufferConfig(_bufferConfig) {
+        if (bufferConfig.save_periodically)
+        {
+          std::thread save_thread(&BufferManager::periodicSave, this);
+          save_thread.detach();
+          //std::thread( [this] { periodicSave(); } );
+        }
     }
-    BufferManager(const std::string& filename, const std::vector<ChannelInfo>& channels,
-                  size_t n_samples, bool auto_save=false) : m_filename(filename), m_auto_save(auto_save), m_n_samples(n_samples){
+    /*BufferManager(const std::string& filename,
+                    const std::vector<ChannelInfo>& channels,
+                    size_t n_samples,
+                    unsigned int _check_period = 10,
+                    size_t _threshold = 0,
+                    bool auto_save=false,
+                    bool _save_periodically = false) : m_filename(filename),
+                  m_auto_save(auto_save), m_n_samples(n_samples), check_period(_check_period),
+                  m_threshold(_threshold), save_periodically(_save_periodically){*/
+    BufferManager(const std::string& filename,
+                    const std::vector<ChannelInfo>& channels,
+                    const BufferConfig& _bufferConfig) : m_filename(filename), bufferConfig(_bufferConfig) {
         assert(!channels.empty());
         assert(!filename.empty());
-        assert(addChannels(channels) == true);
+        auto ret = addChannels(channels);
+        assert(ret == true);
+
+        if (bufferConfig.save_periodically)
+        {
+          std::thread save_thread(&BufferManager::periodicSave, this);
+          save_thread.detach();
+          //std::thread( [this] { periodicSave(); } );
+        }
 	}
 
     ~BufferManager() {
-        if (m_auto_save) {
+        closing = true;
+        if (bufferConfig.m_auto_save) {
             saveToFile();
+        }
+    }
+
+    void periodicSave()
+    {
+        while (!closing)
+        {
+            auto next_step = std::chrono::steady_clock::now() + std::chrono::milliseconds(bufferConfig.check_period);
+
+            // This loop saves all the variables as soon as one of the variables crosses the threshold
+            if (m_buffer_map.size() > 0) // if there are channels
+            {
+                for (auto& [var_name, buff] : m_buffer_map)
+                {
+                    if (buff.size() >= bufferConfig.threshold)
+                    {
+                    saveToFile();
+                    break;
+                    }
+                }
+            }
+            if (std::chrono::steady_clock::now() < next_step)
+            {
+            std::this_thread::sleep_until(next_step);
+            }
         }
     }
 
@@ -58,7 +119,7 @@ public:
 
     bool addChannel(const ChannelInfo& channel) {
         // Probably one day we will have just one map
-        auto ret_buff = m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(channel.m_var_name, Buffer<T>(m_n_samples)));
+        auto ret_buff = m_buffer_map.insert(std::pair<std::string, yarp::telemetry::Buffer<T>>(channel.m_var_name, Buffer<T>(bufferConfig.n_samples)));
         auto ret_dim =  m_dimensions_map.insert(std::pair<std::string, yarp::telemetry::dimensions_t>(channel.m_var_name, channel.m_dimensions));
         return ret_buff.second && ret_dim.second;
     }
@@ -156,10 +217,12 @@ public:
             std::cout << "No available data to be saved" << std::endl;
             return false;
         }
-        auto point_pos = m_filename.find('.');
-        matioCpp::Struct timeSeries(std::string(m_filename.begin(), m_filename.begin()+point_pos), signalsVect);
+        matioCpp::Struct timeSeries(m_filename, signalsVect);
         // and finally we write the file
-        matioCpp::File file = matioCpp::File::Create(m_filename);
+        // since we might save several files, we need to index them
+        std::string new_file = m_filename + "_" + std::to_string(file_index) + ".mat";
+        file_index++;
+        matioCpp::File file = matioCpp::File::Create(new_file);
         return file.write(timeSeries);
     }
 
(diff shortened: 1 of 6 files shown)


src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -27,36 +30,94 @@ struct ChannelInfo {
...
-                  size_t n_samples, bool auto_save=false) : m_filename(filename), m_auto_save(auto_save), m_n_samples(n_samples){
+    /*BufferManager(const std::string& filename, 
+                    const std::vector<ChannelInfo>& channels,
+                    size_t n_samples, 
+                    unsigned int _check_period = 10, 
+                    size_t _threshold = 0, 
+                    bool auto_save=false, 
+                    bool _save_periodically = false) : m_filename(filename), 
+                  m_auto_save(auto_save), m_n_samples(n_samples), check_period(_check_period), 
+                  m_threshold(_threshold), save_periodically(_save_periodically){*/
```
Nicogene, line 64: Maybe we may remove the old version

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -153,16 +214,24 @@ class BufferManager {
...
+    bool closing{false};
+    int file_index{0};
     std::string m_filename;
-    bool m_auto_save{false};
-    size_t m_n_samples{0};
+    //bool m_auto_save{false};
+    //bool save_periodically{false};
+    //unsigned int check_period;
+    //size_t m_n_samples{0};
+    //size_t m_threshold{0};
```
Nicogene, line 234: We can remove also these one

Nicogene (changes requested):

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -27,36 +30,94 @@ struct ChannelInfo {
...
 
     ~BufferManager() {
-        if (m_auto_save) {
+        closing = true;
+        if (bufferConfig.m_auto_save) {
             saveToFile();
         }
     }
 
+    void periodicSave()
```
S-Dafarra, line 88: I would suggest this to be ``private`` to avoid the user calling this manually.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -27,36 +30,94 @@ struct ChannelInfo {
...
+                    if (buff.size() >= bufferConfig.threshold)
+                    {
+                    saveToFile();
+                    break;
+                    }
+                }
+            }
+            if (std::chrono::steady_clock::now() < next_step)
+            {
+            std::this_thread::sleep_until(next_step);
```
S-Dafarra, line 108: Indentation issue

Nicogene (approved):

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -153,19 +179,47 @@ class BufferManager {
...
+            auto next_step = std::chrono::steady_clock::now() + std::chrono::milliseconds(bufferConfig.check_period);
+
+            // This loop saves all the variables as soon as one of the variables crosses the threshold
+            if (m_buffer_map.size() > 0) // if there are channels
+            {
+                for (auto& [var_name, buff] : m_buffer_map)
+                {
+                    if (buff.size() >= bufferConfig.threshold)
+                    {
+                        saveToFile();
```
S-Dafarra, line 205: Maybe it is worth adding a mutex in ``saveToFile`` in case the user calls the method at the same time.

Nicogene: While working on it @AlexAntn and I discussed how to handle the concurrency.
As you pointed out the user can call `saveToFile()` while the thread is already executing the `saveToFile()` function and this a possible race condition.
Same thing if `push_back` is invoked and at the same time the `buff.clear()`  inside `saveToFile` is called by the thread spawn.
For sure we need a lock system, the first implementation may be simply using `std::mutex`
In the future we may think to a lock-free implementation, we can discuss how to handle concurrency during the next meeting

S-Dafarra: Sounds good! In any case, I am not sure if it is possible to save two ``mat`` files from two concurrent threads. This is something I have never investigated, so I am not sure if a lock-free implementation is entirely possible.

AlexAntn: I think, with the current implementation, the main issues would arise from 2 points: 

- when the buffer is being cleared (since the other thread might still be reading from the buffer);
- when it is writing the file itself (if both threads are writing the file at the same exact moment);

Both cases could be "fixed" with a mutex in the `saveToFile()` function, the only drawback I could see is some files in the sequence having only a single value (instead of the threshold used in the periodic saving). But then again, the user should never be calling the `saveToFile()` function I think? We can discuss it

S-Dafarra: > But then again, the user should never be calling the `saveToFile()` function I think? We can discuss it

If so, then the option to enable or disable the automatic saving would not make any sense.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -29,36 +32,59 @@ struct ChannelInfo {
     dimensions_t m_dimensions{ 1,1 };
 };
 
+struct BufferConfig {
+    size_t n_samples = 0;
+    unsigned int check_period = 10;
+    size_t threshold = 0;
+    bool m_auto_save = false;
+    bool save_periodically = false;
+};
```
traversaro, line 41: As mentioned in the meeting, adding a single line of description for this parameters may be beneficial, as the meaning of things like `threshold` may not be obvious. Furthermore, I would either use the `m_` prefix for all the parameters, or not use it at all.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -29,36 +32,59 @@ struct ChannelInfo {
...
-    BufferManager(const std::string& filename, const std::vector<ChannelInfo>& channels,
-                  size_t n_samples, bool auto_save=false) : m_filename(filename), m_auto_save(auto_save), m_n_samples(n_samples){
+
+    BufferManager(const std::string& filename,
+                  const std::vector<ChannelInfo>& channels,
+                  const BufferConfig& _bufferConfig) : m_filename(filename), bufferConfig(_bufferConfig) {
         assert(!channels.empty());
         assert(!filename.empty());
-        assert(addChannels(channels) == true);
+        auto ret = addChannels(channels);
```
traversaro, line 62: Personal taste, so feel free to do as you @AlexAntn @Nicogene prefer, but in this case I think that using too much auto is a bit detrimental to readability, and having `bool ret`  may be more clear.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
traversaro: I think that disabling autosave to avoid that any thread is created may be useful in some cases. On the other hand, we can simply clarify that if autosave is enabled, `saveToFile`  does not works and it will return `false` .

Nicogene:
I was thinking to merge this PR as it is and open soon another PR addressing the remaining changes required + more improvements.

The upcoming PR will break probably the API once again 😅 

cc @GiulioRomualdi @S-Dafarra @traversaro @AlexAntn

Issue #56, opened 2021-02-01
Discuss the possibility to remove YARP dependency from telemetery


GiulioRomualdi:
Hi all, I was checking yarp telemetry. It would be great to embed it in [`bipedal-locomotion-framework`](https://github.com/dic-iit/bipedal-locomotion-framework) (aka `blf`). However, many libraries in `blf` do not depend on `YARP`. 

As far as I understood the only reason why `YARP` is used to have all the data synchronized with the same clock.

https://github.com/robotology-playground/yarp-telemetry/blob/306c9a047655fc65bf8072006193291bde767346/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h#L75-L85

I theory this dependency can be easily dropped by passing a function to generate the clock. Thanks to this approach different kinds of clocks can be used. For instance `yarp clock`, `std clock`, `ros clock`).

The following patch implements what I proposed https://github.com/robotology-playground/yarp-telemetry/compare/master...GiulioRomualdi:remove/yarp_dependency

Looking forward to receiving your  feedback :smile: 

cc @traversaro @DanielePucci @S-Dafarra @Nicogene @AlexAntn

Nicogene:
I personally like very much your implementation, besides removing the yarp dependency.

For now, yarp is used only for the clock but when the complexity will increase (e.g. handle and collect recording from different machines), yarp may become indispensable 

In any case, a PR with the changes you proposed is more than welcome!

PR #58, opened 2021-02-01
Remove yarp dependency from yarp telemetry

GiulioRomualdi:
This follows #56 

cc @Nicogene

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -15,8 +15,9 @@
 #include <vector>
 #include <iostream>
 #include <assert.h>
-#include <yarp/os/Time.h>
+#include <functional>
 #include <matioCpp/matioCpp.h>
+
 
 namespace yarp::telemetry {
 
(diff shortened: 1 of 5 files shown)


Nicogene (commented):
Thanks for submitting the PR!
We will handle this after discussing and merging #55 

BTW we may need a default value `m_nowFunction` or handle the case in which the user forgot to set the now function

GiulioRomualdi:
> BTW we may need a default value m_nowFunction or handle the case in which the user forgot to set the now function

I can deal with that

GiulioRomualdi:
Done in https://github.com/robotology-playground/yarp-telemetry/pull/58/commits/195b96b7f9f68c623fd2150588e201d3376e302f

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -159,13 +162,29 @@ class BufferManager {
...
+            return false;
+        }
+
+        m_nowFunction = now;
+        return true;
+    }
+
 private:
+    static double DefaultClock() {
+        return std::chrono::duration<double>(std::chrono::system_clock::now().time_since_epoch()).count();
```
traversaro, line 179: Are you sure that this is converting the time in **seconds** since the epoch? @GiulioRomualdi

Nicogene: I tried this code snippet
```c++
#include <iostream>
#include <chrono>
#include <iomanip>
#include <unistd.h>

int main()
{
    auto now = std::chrono::duration<double>(std::chrono::system_clock::now().time_since_epoch()).count();
    sleep(1);
    auto then = std::chrono::duration<double>(std::chrono::system_clock::now().time_since_epoch()).count();
    std::cout<<"time: "<<then - now<<std::endl;
    return 0;
}

```
and I get this output
```
time: 1.00014 
```

So it should be in seconds according to [this](https://pubs.opengroup.org/onlinepubs/009695399/functions/sleep.html)

Nicogene: See also [YARP implementation](https://github.com/robotology/yarp/blob/9b5ebc9ccce2d03ccc49b6aa2e7469c4cba710b4/src/libYARP_os/src/yarp/os/SystemClock.cpp#L37-L40)

Nicogene (approved):

PR #65, opened 2021-02-05
BufferManager: add mutex, configuration from file, empty constructor workflow et al

### LINKED ISSUE #39: Investigation file-based configuration of the buffer manager

### LINKED ISSUE #60: Support use of the default argument-less constructor for BufferManager class
This could be useful also for https://github.com/robotology-playground/yarp-telemetry/issues/39 .

### LINKED ISSUE #61: Enforce relative condition between threshold (the max number of samples before) and number of samples

### LINKED ISSUE #66: json parser: use fetch content
In #65 we added the json.hpp in the repo but it would be better to use cmake FetchContent

### LINKED ISSUE #68: Add support to save the configuration

Nicogene:
- The `ChannelInfo` has been changed from a struct of two elements to a
`std::pair`, in order to be easily handled with the JSON file conf.
- All the flags, options of the BufferManager have been put in the BufferConfig
- The empty constructor is allowed now, for this we added all the functions
needed for configuring a default BufferManager(It fixes #60)

- The mutex has been added in `saveToFile`(partially address #9)
- Now the periodicSave, saves only the variables that go above the
data_treshold specified
- Fixes the bug of not saving when the threshold > n_samples(fixes #61 )
- It has been added the configuration from JSON file using [this library](https://github.com/nlohmann/json) with relative example (it fixes #39) 
- It add the possibility to save `BufferConfig` to json file(it fixes #68)
- It add `nlhomann_json v3.9.0` as dependency via `FetchContent`( it fixes #66)
Here is an example of JSON file for configuring the `BufferManager`:

```json
{
    "filename": "buffer_manager_test_conf_file",
    "n_samples": 20,
    "save_period": 1.0,
    "data_threshold": 10,
    "auto_save": true,
    "save_periodically": true,
    "channels": [
        ["one",[1,1]],
        ["two",[1,1]]
    ]
  }

```

For now we used the json library just importing the `json.hpp` file, this has to be in the future to an "extern" folder or better with the cmake FetchContent mechanism.

Please review code.

cc @traversaro @S-Dafarra @GiulioRomualdi

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,24 @@ find_package(matioCpp REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Threads REQUIRED)
 
+option(YARP_TELEMETRY_USES_SYSTEM_nlohmann_json OFF)
+# 3.9.2 is unreleased, this option is not working until that version is not released
+if(YARP_TELEMETRY_USES_SYSTEM_nlohmann_json)
+  find_package(nlohmann_json 3.9.2 REQUIRED)
+else()
+  include(FetchContent)
+
+  FetchContent_Declare(json
+    GIT_REPOSITORY https://github.com/nlohmann/json.git
+    GIT_TAG develop) # awaiting 3.9.2 release for https://github.com/nlohmann/json/issues/2513
+
+  FetchContent_GetProperties(json)
+  if(NOT json_POPULATED)
+    FetchContent_Populate(json)
+    add_subdirectory(${json_SOURCE_DIR} ${json_BINARY_DIR} EXCLUDE_FROM_ALL)
+  endif()
+endif()
+
 
 feature_summary(WHAT ALL INCLUDE_QUIET_PACKAGES)
 
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -79,6 +79,11 @@ public:
 
     void setFileName(const std::string& filename) {
         m_filename = filename;
+    void resize(size_t new_size) {
+        for (auto& [var_name, buff] : m_buffer_map) {
+            buff.resize(new_size);
+        }
+        m_bufferConfig.n_samples = new_size;
         return;
     }
 
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -27,17 +27,16 @@ namespace yarp::telemetry {
 
 using dimensions_t = std::vector<size_t>;
 
-struct ChannelInfo {
-    std::string m_var_name;
-    dimensions_t m_dimensions{ 1,1 };
-};
+using ChannelInfo = std::pair< std::string, dimensions_t >;
 
 struct BufferConfig {
-    size_t n_samples = 0;
-    unsigned int check_period = 10;
-    size_t threshold = 0;
-    bool m_auto_save = false;
-    bool save_periodically = false;
+    std::string filename{ "" };
+    size_t n_samples{ 0 };
+    double save_period{ 0.010 };
+    size_t data_threshold{ 0 };
+    bool auto_save{ false };
+    bool save_periodically{ false };
+    std::vector<ChannelInfo> channels;
 };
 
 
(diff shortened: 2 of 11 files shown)


src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -10,82 +10,110 @@
...
+    bool configure(const BufferConfig& _bufferConfig) {
+        bool ok{ true };
+        m_bufferConfig = _bufferConfig;
+        if (!_bufferConfig.channels.empty()) {
+            ok = ok && addChannels(_bufferConfig.channels);
+        }
+        if (ok && _bufferConfig.save_periodically) {
+            ok = ok && enablePeriodicSave(_bufferConfig.save_period);
+        }
+        // TODO ROLL BACK IN CASE OF FAILURE
```
traversaro, line 94: If you do not plan to address these, please open issues for any open points. TODO in the code get forgotten quite easily, better to open issues to track then.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -10,82 +10,110 @@
...
+        }
+        return false;
+    }
+    bool configureFromFile(const std::string& config_filename) {
+        // read a JSON file
+        std::ifstream input_stream(config_filename);
+        nlohmann::json jason_file;
+        input_stream >> jason_file;
+        return configure(jason_file.get<yarp::telemetry::BufferConfig>());
+    }
```
traversaro, line 84: Can't we have this as a free function instead, something like `bool configureFromJSONFile(BufferManager& manager, const std::string& config_filename)` ? 
This would avoid needles coupling of the specific configuration method with the BufferManager class.

Nicogene:
Before merging this maybe we should address #66 that seems straightforward and in this way, we avoid having so many additions

CMakeLists.txt
```diff
@@ -62,6 +62,23 @@ find_package(matioCpp REQUIRED)
 find_package(Boost REQUIRED)
 find_package(Threads REQUIRED)
 
+option(YARP_TELEMETRY_USES_SYSTEM_nlohmann_json OFF)
+if(YARP_TELEMETRY_USES_SYSTEM_nlohmann_json)
+  find_package(nlohmann_json 3.9.0 REQUIRED)
```
traversaro, line 67: Do you really require 3.9? Ubuntu 20.04 ships with 3.7 in apt (https://repology.org/project/nlohmann-json/versions), I would try to at least enable it, if then it does not work we can fix it to the minimum working version.

Nicogene: We need 3.9.0 for [this macro](https://github.com/nlohmann/json/blob/bb90e34d22b59ad3a15e459000fbe5a693b32397/single_include/nlohmann/json.hpp#L2434-L2445). But we can write the functions by hand.

Nicogene:
Looking https://github.com/robotology-playground/yarp-telemetry/pull/65/checks?check_run_id=1853547617 seems that someone is putting `BUILD_TESTING` to true

Nicogene:
Here is why https://github.com/nlohmann/json/issues/2513, we need 3.9.2

CMakeLists.txt
traversaro: If we actually need 3.9 then it is ok for me!

traversaro:
> 
> 
> Here is why [nlohmann/json#2513](https://github.com/nlohmann/json/issues/2513), we need 3.9.2(that is unreleased)

FetchContent can also use a specific commit, and then we can switch to 3.9.2 once is released.

Nicogene:
> FetchContent can also use a specific commit, and then we can switch to 3.9.2 once is released.

I put `develop` branch

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
Nicogene: This has been addressed in https://github.com/robotology-playground/yarp-telemetry/pull/65/commits/b5b431ea6f05f5a0bcda116206d7c71147d86233.

Instead of `BufferManager` I put as I/O `BufferConfig`, in order to avoid circular dependencies

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
Nicogene: Issue #72 opened

traversaro (approved):

PR #77, opened 2021-02-15
Add timestamp into the filename

### LINKED ISSUE #63: Cleanup how periodic save is data (naming)

Nicogene:
It fixes #63 

Here is the resulting filename

`buffer_manager_test_conf_file_1613379268.571470.mat`

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -291,8 +291,7 @@ public:
         matioCpp::Struct timeSeries(m_bufferConfig.filename, signalsVect);
         // and finally we write the file
         // since we might save several files, we need to index them
-        std::string new_file = m_bufferConfig.filename + "_" + std::to_string(file_index) + ".mat";
-        file_index++;
+        std::string new_file = m_bufferConfig.filename + "_" + std::to_string(m_nowFunction()) + ".mat";
         matioCpp::File file = matioCpp::File::Create(new_file);
         return file.write(timeSeries);
     }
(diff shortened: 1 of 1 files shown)


Nicogene:
I found out also the ISO 8601 format https://stackoverflow.com/questions/36927626/outputting-date-in-iso-8601-format as @traversaro suggested

traversaro:
What is the plan? Switch to ISO  8601  or merge this as it is and improve it later? My experience with yarplogger and other software that use timestamp in seconds to present time information to users may be confusing (it was a typical complain from @francesco-romano back in time, for example).

Nicogene:
> What is the plan? Switch to ISO 8601 or merge this as it is and improve it later? 

We can discuss it, the solution proposed in the link I put relies on `std::chrono` but our API gives the possibility to set a "custom" now function. I don't know if we should use that now function or for the naming of the file we can use always std::chrono.

traversaro:
> 
> 
> > What is the plan? Switch to ISO 8601 or merge this as it is and improve it later?
> 
> We can discuss it at the next meeting, the solution proposed in the link I put relies on `std::chrono` but our API gives the possibility to set a "custom" now function. I don't know if we should use that now function or for the naming of the file we can use always std::chrono.

We can discuss in the meeting, but I think we should not mix two different aspects: 
* The possibility of setting a custom now 
* How to "translate" in human-friendly form the time instant provided by any now

The second aspect should be the same regardless of which now function is used.

Nicogene:
We discussed the pros and cons of both solutions, sure  ISO 8601 makes the date and time of logging more human-readable, but it is not coherent with the format in which is saved the timestamp for the logged variables.
Moreover if using a custom clock(e.g. gazebo ones) that starts from 0, the conversion to ISO 8601 will give an odd date and time(e.g. files dated in the 70s).

For "fixing" this we may specify in the documentation that the `nowFunction` should be a function that returns a double representing the seconds elapsed since the epoch.

Anyhow, merging as it is, is the simplest solution and we may change it if needed by the customer.

PR #78, opened 2021-02-15
Add first draft of unit tests

### LINKED ISSUE #54: Add unit test in the CI
We are in a stage where the API change often so it is difficult to implement and maintain unit tests, but at some point, we should introduce unit tests in the CI.

My proposal is to use [Catch2](https://github.com/catchorg/Catch2)

Nicogene:
It fixes #54 

This PR add the first test case `BufferManagerTest` that tests all the functionalities we added so far to the `BufferMananger` so far.

There are still open points e.g:

- I had to add delays in order to avoid segfaults due to the destruction of `BufferManager` object too soon(#9 is not completely fixed, we have to investigate about it cc @AlexAntn )
~- Locally I had issues with `nlohmann_json`, enabling the tests enabled also the json unit tests, I want to see if this problem is happening also in the CI.~
This seems to not happen if using a clean workspace with `develop` branches of `nlohmann_json`
- Find a way to check if something has been saved in the case of the periodic save, and it is something valid.


Nicogene:
The issue about concurrency does not happen always, here what happens:

```
Test project C:/Users/ngenesio/robotology/yarp-telemetry/build/x64-Debug
    Start 62: Buffer Manager Test
Process not started
 C:/Users/ngenesio/robotology/yarp-telemetry/build/x64-Debug/bin/bufferManager_test.exe
[resource busy or locked]
1/1 Test #62: Buffer Manager Test ..............***Not Run   0.00 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.04 sec

The following tests FAILED:
         62 - Buffer Manager Test (BAD_COMMAND)
Errors while running CTest

```

Nicogene:
Removing the delays the problem occurs:
https://github.com/robotology-playground/yarp-telemetry/runs/1904540827?check_suite_focus=true

Moreover the test seems to not be run under windows, we need to investigate it

Nicogene:
This https://github.com/robotology-playground/yarp-telemetry/pull/78/commits/f7e5014205607f77ee0da916206eeced7d171e9c seemed to fix the race condition but under `macOs` seems to fail again.

I tried on my pc(Windows) this
` ctest --repeat-until-fail 100 `

and never failed.

It is failing the last test that enables the periodic save, but not the auto_save.

traversaro:
If you need to links the threads library, it should be sufficient to link privately to the `Threads::Threads` imported target.

Nicogene:
This 
> - Find a way to check if something has been saved in the case of the periodic save, and it is something valid.

Will be addressed in a second step/PR for improving the tests. We will use matio-cpp for reading the saved file.

And this seemed to affect only `macOs` and now should be fixed
> - I had to add delays in order to avoid segfaults due to the destruction of `BufferManager` object too soon

AlexAntn (approved):

traversaro (approved):

Issue #15, opened 2020-09-16
yarp-device-dumper: Proposal for a YARP device-based logger to run in yarprobotinterface


traversaro:
**Note: this is not meant to be a replacement for full telemetry, but a small prototype can be easily implemented in a short range of time 
to achieve the goal of having a dumper always running to help debug of failures.**

# `yarp-device-dumper`

Similar to how the `yarpdatadumper` is a process used to connect to a YARP port and save the data on a file,
the `yarp-device-dumper` is a YARP device that is meant to run as part of a `yarprobotinterface` instance,
that runs a periodic thread that read data from a series of YARP devices passed via the `attach` action of the  `yarprobotinterface`.
For each thread run, the data is added in a [circular buffer](https://en.wikipedia.org/wiki/Circular_buffer), so that the 
device can always run without the risk of filling the `icub-head` PC memory. The size of the circular buffer can be 
configured to keep in memory only the desired last X minutes of experiments, to have a tradeoff between the data that 
can be saved and the use of memory. 

Once the yarprobotinterface is closed, or a write to file is requested in another way (such as a periodic save or a a RPC call), the 
data from the circular buffer is dumped to a `.mat` file on disk. 

By design, the data that is possible to collect will be configurable, but not arbitrary: the data can be collected only if the device 
expose them in properly supported YARP interfaces. For a first prototype, I guess it will be important to support control board interfaces,
such as [`ITorqueControl`](https://www.yarp.it/classyarp_1_1dev_1_1ITorqueControl.html) or [`IPositionControl`](https://www.yarp.it/classyarp_1_1dev_1_1IPositionControl.html).
For a second iteration, supporting also the data published on Multiple Analog Sensors interface. 

Main advantages over the existing pattern of having multiple yarpdatadumpers for all the ports that needs to be saved:
* No need to start manually the dumper: the dumper starts automatically as soon as you start the robot.
* No need to worry about clock time the data: data from different robot parts or from different sensors
  is implicitly synchronized by the fact that are read at the same time.
* No need to parse and merge multiple textual files: the data is already in a `.mat` file, 
  ready for plotting/post processing in MATLAB/Python/Julia/C++. 
* No need to worry about disconnection between the icub-head PC and the laptop on which yarpdatadumper
  are running (useful for when the iCub is running on Wi-Fi).
* No need to explicitly enable it: thanks to the use of the circular buffer, the dumper can just run always,
[...]

PR #84, opened 2021-02-26
telemetryDeviceDumper: add first draft

Nicogene:
This addresses #15 

I made it `IWrapper`/`IMultipleWrapper` but we can discuss the use cases (cc @AlexAntn @S-Dafarra @traversaro )

@AlexAntn and I implemented the first working version

For now it dumps the encoders, the encoders' speed and accelerations for the specified parts of the specified robot.
Here is an example of usage:

```bash
$ yarpdev --device telemetryDeviceDumper --parts "(head left_arm right_leg)"
```
(if not specified `icubSim` is the default value for `robot`)

And here are the file produced:
```
-a----          3/1/2021   3:20 PM          55696 telemetryDeviceDumper_icubSim_head_1614608427.576625.mat
-a----          3/1/2021   3:20 PM          62584 telemetryDeviceDumper_icubSim_head_1614608431.579932.mat
-a----          3/1/2021   3:20 PM          62592 telemetryDeviceDumper_icubSim_head_1614608435.584401.mat
-a----          3/1/2021   3:20 PM          61184 telemetryDeviceDumper_icubSim_head_1614608439.589465.mat
-a----          3/1/2021   3:20 PM         141784 telemetryDeviceDumper_icubSim_left_arm_1614608427.805764.mat
-a----          3/1/2021   3:20 PM         149944 telemetryDeviceDumper_icubSim_left_arm_1614608431.807887.mat
-a----          3/1/2021   3:20 PM         149944 telemetryDeviceDumper_icubSim_left_arm_1614608435.811885.mat
-a----          3/1/2021   3:20 PM          62584 telemetryDeviceDumper_icubSim_right_leg_1614608428.033400.mat
-a----          3/1/2021   3:20 PM          62584 telemetryDeviceDumper_icubSim_right_leg_1614608432.036649.mat
-a----          3/1/2021   3:20 PM          62584 telemetryDeviceDumper_icubSim_right_leg_1614608436.041214.mat
```

and here it is the content of `telemetryDeviceDumper_icubSim_head` for example:
```

telemetryDeviceDumper_icubSim_head = 

  struct with fields:

                 encoders: [1×1 struct]
           encoder_speeds: [1×1 struct]
    encoder_accelerations: [1×1 struct]

telemetryDeviceDumper_icubSim_head.encoders =

  struct with fields:

          data: [1×6×356 double]
    dimensions: [1 6 356]
          name: 'encoders'
    timestamps: [1×356 double]

```

And so on.

I think we achieved the first milestone, improvements can be made(e.g. the configuration of the buffermanager is hardcoded, we can use the configuration from json)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ endif()
 
 
 set(YARP_FORCE_DYNAMIC_PLUGINS TRUE CACHE INTERNAL "${PROJECT_NAME} is always built with dynamic plugins")
-find_package(YARP 3.4 COMPONENTS conf os REQUIRED)
+find_package(YARP 3.4 COMPONENTS conf dev os REQUIRED)
 # Use the same build options as YARP
 set(BUILD_SHARED_LIBS ${YARP_IS_SHARED_LIBRARY})
 set(CMAKE_C_FLAGS ${YARP_C_FLAGS})
--- a/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
+++ b/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2006-2021 Istituto Italiano di Tecnologia (IIT)
+ * All rights reserved.
+ *
+ * This software may be modified and distributed under the terms of the
+ * BSD-3-Clause license. See the accompanying LICENSE file for details.
+ */
+#include "TelemetryDeviceDumper.h"
+
+using namespace yarp::telemetry;
+
+TelemetryDeviceDumper::TelemetryDeviceDumper() : yarp::os::PeriodicThread(0) {
+    // TODO
+}
+
+TelemetryDeviceDumper::~TelemetryDeviceDumper() {
+    // TODO
+}
+
+bool TelemetryDeviceDumper::close() {
+    // TODO
+    return true;
+}
+
+bool TelemetryDeviceDumper::open(yarp::os::Searchable& config) {
+    // TODO
+    return true;
+}
+
+bool TelemetryDeviceDumper::attachAll(const yarp::dev::PolyDriverList& device2attach) {
+    // TODO
+    return true;
+}
+
+bool TelemetryDeviceDumper::detachAll() {
+    // TODO
+    return true;
+}
+
+bool TelemetryDeviceDumper::attach(yarp::dev::PolyDriver* poly) {
+    // TODO
+    return true;
+}
+
+bool TelemetryDeviceDumper::detach() {
+    // TODO
+    return true;
+}
+
+// PeriodicThread
+bool TelemetryDeviceDumper::threadInit() {
+    // TODO
+    return true;
+}
+
+void TelemetryDeviceDumper::threadRelease() {
+    // TODO
+    return;
+}
+
+void TelemetryDeviceDumper::run() {
+    // TODO
+    return;
+}
+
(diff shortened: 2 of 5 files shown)


Nicogene:
I made this PR ready for review and fill in the body of the PR with more details.

cc @S-Dafarra, @GiulioRomualdi, @pattacini, @drdanz, @vtikha @traversaro

CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ endif()
 
 
 set(YARP_FORCE_DYNAMIC_PLUGINS TRUE CACHE INTERNAL "${PROJECT_NAME} is always built with dynamic plugins")
-find_package(YARP 3.4 COMPONENTS conf os REQUIRED)
+find_package(YARP 3.4 COMPONENTS conf dev os REQUIRED)
```
traversaro, line 57: I thought at some point we remove the hard dependency on YARP, was not that the case @GiulioRomualdi ? Could it make sense that we have a `YARP_TELEMETRY_USES_YARP_OS`  option that can be disable to avoid compiling the part that depend on YARP_os like the `telemetryDeviceDumper` ?

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+using namespace yarp::telemetry;
+using namespace yarp::os;
+
+constexpr double period_thread{ 0.010 };
+constexpr std::array<char*, 6> partsArray{ "head","torso","left_arm","right_arm","left_leg","right_leg" };
+
+
+bool RemoteControlGateway::open(const std::string& robot, const std::string& part, const std::string& moduleName) {
+    yarp::os::Property conf{ {"device", yarp::os::Value("remote_controlboard")},
```
traversaro, line 19: I don't think it make sense to access the controlboard functionality via YARP network if the device is already running as part of the yarprobotinterface. Instead, the `telemetryDeviceDumper`  should receive the device from which to read in the `attachAll` method, and read from them without opening any port. This also minimizes the delay in data read.

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,158 @@
...
+    }
+    PeriodicThread::stop();
+    return true;
+}
+
+bool TelemetryDeviceDumper::open(yarp::os::Searchable& config) {
+    auto robot = config.check("robot", Value("icubSim")).asString();
+    auto name = config.check("name", Value("telemetryDeviceDumper")).asString();
+
+    auto partsBot = config.find("parts").asList();
```
traversaro, line 97: As mentioned in https://github.com/robotology-playground/yarp-telemetry/pull/84/files#r585418226, the device from which to connect should be received from attachAll rather then be opened here. This should also remove the need for a configuration for this.

traversaro:
I have some doubt related to how this is different w.r.t. to the proposal in https://github.com/robotology-playground/yarp-telemetry/issues/15, in particular w.r.t to:
> Similar to how the yarpdatadumper is a process used to connect to a YARP port and save the data on a file,
the yarp-device-dumper is a YARP device that is meant to run as part of a yarprobotinterface instance,
that runs a periodic thread that read data from a series of YARP devices passed via the attach action of the yarprobotinterface.

It seems to me that this version has quite similar shortcomings to the use of yarpdatadumper, but we can discuss about it in the meeting.

traversaro:
As the required improvements will be tracked in https://github.com/robotology-playground/yarp-telemetry/issues/85, we can merge this PR once comment https://github.com/robotology-playground/yarp-telemetry/pull/84#discussion_r585416311 is sorted.

traversaro:
I guess we can merge, we can always modify the CMake conf further once needed by @GiulioRomualdi .

Issue #88, opened 2021-03-02
telemetryDeviceDumper: support specify the directory in which log files are saved


traversaro:

traversaro:
@AlexAntn this will probably required a new option/method in libYARP_telemetry

Issue #86, opened 2021-03-02
Add to telemetryDeviceDumper  the possibility of saving the names of the joints in a cell array of string (or equivalent method)


traversaro:
Joint names can be obtained via  http://www.yarp.it/git-master/classyarp_1_1dev_1_1IAxisInfo.html#a1350180d6aaaba2dd8f22c42655f207e .

traversaro:
We could have an info structure for each channel.

PR #89, opened 2021-03-04
 telemetryDeviceDumper: first draft attaching multiple controlboards and use the remapper

### LINKED ISSUE #85: Transform telemetryDeviceDumper to attach to devices via IMultipleWrapper
Regarding the data that we can save, we can take as inspiration the options supported by https://robotology.github.io/robotology-documentation/doc/html/group__icub__controlBoardDumper.html .

Nicogene:
It fixes #85 

It has to be tested and it misses some API features of the library for work correctly
(see #86 and #88 )

It has to be merged after #84 

cc @AlexAntn @traversaro @S-Dafarra @drdanz

### DIFF
--- a/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
+++ b/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
@@ -101,7 +101,7 @@ bool TelemetryDeviceDumper::loadSettingsFromConfig(yarp::os::Searchable& config)
 }
 
 bool TelemetryDeviceDumper::open(yarp::os::Searchable& config) {
-    
+
     std::lock_guard<std::mutex> guard(this->deviceMutex);
 
     bool ok;
@@ -206,13 +206,13 @@ bool TelemetryDeviceDumper::configBufferManager(yarp::os::Searchable& conf) {
     // This is the buffer manager configuration
     yarp::telemetry::BufferConfig bufferConfig;
     bool ok{ true };
-    ok = ok && bufferManager.addChannel({ "encoders", {1, jointPos.size()} });
+    ok = ok && bufferManager.addChannel({ "encoders", {jointPos.size(), 1} });
     if (ok && settings.logJointVelocity) {
-        ok = ok && bufferManager.addChannel({ "velocity", {1, jointVel.size()} });
+        ok = ok && bufferManager.addChannel({ "velocity", {jointVel.size(), 1} });
     }
 
     if (ok && settings.logJointAcceleration) {
-        ok = ok && bufferManager.addChannel({ "acceleration", {1, jointAcc.size()} });
+        ok = ok && bufferManager.addChannel({ "acceleration", {jointAcc.size(), 1} });
     }
 
     // TODO: some settings from the ini are duplicated respect to the settings specified in the json file
--- a/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
+++ b/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
@@ -7,70 +7,54 @@
  */
 #include "TelemetryDeviceDumper.h"
 #include <array>
+#include <algorithm>
+#include <cmath>
 
 using namespace yarp::telemetry;
 using namespace yarp::os;
+using namespace yarp::dev;
 
 constexpr double period_thread{ 0.010 };
-constexpr std::array<char*, 6> partsArray{ "head","torso","left_arm","right_arm","left_leg","right_leg" };
 
 
-bool RemoteControlGateway::open(const std::string& robot, const std::string& part, const std::string& moduleName) {
-    yarp::os::Property conf{ {"device", yarp::os::Value("remote_controlboard")},
-                             {"remote", yarp::os::Value("/" + robot + "/" + part)},
-                             {"local",  yarp::os::Value("/" + moduleName + "/" + part + "/remoteControlBoard")} };
+void convertVectorFromDegreesToRadians(std::vector<double>& inVec) {
+    std::transform(inVec.begin(), inVec.end(), inVec.begin(), [](double el) -> double { return el * M_PI / 180.0; });
+}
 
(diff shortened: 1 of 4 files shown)


src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,331 @@
...
+        else
+        {
+            bufferManager.push_back(jointPos, "acceleration");
+        }
+
+
+    }
+    if (settings.useRadians) {
+        // TODO check if it is safe to call transform on empty vectors.
+        convertVectorFromDegreesToRadians(jointPos);
```
traversaro, line 317: Joint in YARP can be either revolute or prismatic, you can get this info from the IAxisInfo interface.

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+bool TelemetryDeviceDumper::openRemapperControlBoard(os::Searchable& config)
+{
+    // Pass to the remapper just the relevant parameters (axesList)
+    yarp::os::Property propRemapper;
+    propRemapper.put("device", "controlboardremapper");
+    bool ok = getUsedDOFsList(config, jointNames);
+    if (!ok) return false;
+
+    addVectorOfStringToProperty(propRemapper, "axesNames", jointNames);
```
traversaro, line 138: Probably future improvements, but I think it would be convenient to have the possibility to specify "all joints", especially given this is just a dumper (different is the situation in which the joints are controlled such in the case of the controller, in that case it make more sense to be explicit on which joint to control).

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+bool TelemetryDeviceDumper::attachAllControlBoards(const yarp::dev::PolyDriverList& p_list) {
+    PolyDriverList controlBoardList;
+    for (size_t devIdx = 0; devIdx < (size_t)p_list.size(); devIdx++)
+    {
+        yarp::dev::IEncoders* pEncs = 0;
+        if (p_list[devIdx]->poly->view(pEncs))
+        {
+            controlBoardList.push(const_cast<PolyDriverDescriptor&>(*p_list[devIdx]));
+        }
```
AlexAntn, line 178: We should probably introduce a warning here if a board is not present/active (and the user expects it to be)

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+bool TelemetryDeviceDumper::detachAll()
+{
+    std::lock_guard<std::mutex> guard(this->deviceMutex);
+    correctlyConfigured = false;
+    if (isRunning())
+    {
+        stop();
+    }
+    return  this->remappedControlBoardInterfaces.multwrap->detachAll();;
```
AlexAntn, line 252: double ; at the end of the line

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -0,0 +1,331 @@
...
+    {
+        bufferManager.push_back(jointPos, "encoders");
+    }
+
+    // At the moment we are assuming that all joints are revolute
+
+    if (settings.logJointVelocity)
+    {
+        ok = remappedControlBoardInterfaces.encs->getEncoderSpeeds(jointVel.data());
+        sensorsReadCorrectly = sensorsReadCorrectly && ok;
```
AlexAntn, line 288: shouldn't this be `ok = ok && sensorsReadCorrectly;` ?

(same in line 303)

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
Nicogene: I think that it is already done by the remapper used after, if we ask for example "head, left_arm" to the remapper and we don't pass the correct devices it will fail the open, am I right @traversaro ?

traversaro: Yes, if the axesList contains names that are not exposed by the controlboard wrapper the attachAll of the controlboard will fail.

Nicogene:
In https://github.com/robotology-playground/yarp-telemetry/pull/89/commits/cf5f32f8923bced5b8ca1a651827812c4b7d5adf I changed the configuration of the BM for the tests on the real robot.

On my fork I pushed the XML files for launching the device (https://github.com/Nicogene/robots-configuration/commit/16748c7cd886899b79ceab3257da42eb9c9aadce) with `iCubErzelli02`

cc @AlexAntn

Nicogene:
After some fixes (See #93 ) this is actually the very first working version tested with the real robot launched via `yarprobotinterface`.

See https://github.com/robotology-playground/yarp-telemetry/issues/87#issuecomment-797433463 for details.

I think it is ready for the review

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
Nicogene: Opened #98 for tracking this missing feature

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
Nicogene: Opened #99 for handling this missing feature

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
Nicogene: Actually, now that we read just from the control boards it is equivalent

Nicogene:
As discussed, the device is working and we opened all the issues for handling the missing features and the bugs we spotted. We can merge it as first draft

PR #91, opened 2021-03-05
Add description list and path to the BufferManager configuration

### LINKED ISSUE #86: Add to telemetryDeviceDumper  the possibility of saving the names of the joints in a cell array of string (or equivalent method)

### LINKED ISSUE #88: telemetryDeviceDumper: support specify the directory in which log files are saved

Nicogene:
It is now possible to specify a custom path(`path`) and a list of string(`description_list`) for describing the data logged.

```
>> buffer_manager_test_conf_file = 

  struct with fields:

    description_list: {2×1 cell}
                 one: [1×1 struct]
                 two: [1×1 struct]

>> buffer_manager_test_conf_file.description_list

ans =

  2×1 cell array

    {'This is a test'}
    {'Or it isn't?'  }
```

It fixes #86 
It fixes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ Here is the code snippet for dumping in a `.mat` file 3 samples of the 2x3 matri
     bm_m.resize(3);
     bm_m.setFileName("buffer_manager_test_matrix");
     bm_m.enablePeriodicSave(0.1); // This will try to save a file each 0.1 sec
+    bm_m.setDefaultPath("/my/preferred/path");
+    bm_m.setDescriptionList({"head", "left_arm"});
     std::vector<yarp::telemetry::ChannelInfo> vars{ { "one",{2,3} },
                                                     { "two",{3,2} } };
 
@@ -208,6 +210,9 @@ It is possible to load the configuration of a BufferManager **from a json file**
 Where the file has to have this format:
 ```json
 {
+    "description_list": ["This is a test",
+                         "Or it isn't?"],
+    "path":"/my/preferred/path",
     "filename": "buffer_manager_test_conf_file",
     "n_samples": 20,
     "save_period": 1.0,
@@ -218,7 +223,7 @@ Where the file has to have this format:
         ["one",[1,1]],
         ["two",[1,1]]
     ]
-  }
+}
 ```
 The configuration can be saved **to a json file**
 ```c++
--- a/src/examples/conf/test_json.json
+++ b/src/examples/conf/test_json.json
@@ -1,4 +1,6 @@
 {
+    "description_list": ["This is a test",
+                         "Or it isn't?"],
     "path":"",
     "filename": "buffer_manager_test_conf_file",
     "n_samples": 20,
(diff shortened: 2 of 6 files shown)


Nicogene:
Side note, I noticed that a `description_list` with one empty string cannot be converted to the CellArray

```
[ERROR][matioCpp::Variable::createVar] The name should not be empty.
[ERROR][matioCpp::CellArray::CellArray] The element at index 0 (0-based) is not valid.
Assertion failed: false, file ..\..\..\src\CellArray.cpp, line 81
```

This is because it interprets the string as the name of the variable, it is supposed to be like this @S-Dafarra ?

S-Dafarra:
> Side note, I noticed that a `description_list` with one empty string cannot be converted to the CellArray
> 
> ```
> [ERROR][matioCpp::Variable::createVar] The name should not be empty.
> [ERROR][matioCpp::CellArray::CellArray] The element at index 0 (0-based) is not valid.
> Assertion failed: false, file ..\..\..\src\CellArray.cpp, line 81
> ```
> 
> This is because it interprets the string as the name of the variable, it is supposed to be like this @S-Dafarra ?

Yes, every variable is supposed to have a name. In case of strings, it is possible to consider the name also as the content of the string. This is helpful especially with containers like ``CellArray`` were the name of the elements composing the array is not used. A simple fix is to give the string a name and empty content, i.e ``String("useless_name", "")``

Nicogene:
Ok but in this way we would have something like

```
  2×1 cell array

    {'useless_name','This is a test'}
    {'useless_name','Or it isn't?'  }
```

I'd prefer the actual version, now I am skipping the empty strings.
What do you think about it?

S-Dafarra:
> Ok but in this way we would have something like
> 
> ```
>   2×1 cell array
> 
>     {'useless_name','This is a test'}
>     {'useless_name','Or it isn't?'  }
> ```
> 
> I'd prefer the actual version, now I am skipping the empty strings.
> What do you think about it?

No, the name of the elements is not visible in the cell_arrays. For example, this is how the ``cell_array`` variable looks like 
image

when written with the code of https://github.com/dic-iit/matio-cpp/blob/0f1cc62534f331c5db6e61b89c3ce7d071310567/test/FileUnitTest.cpp#L298-L307

As you can see, the names disappear.

It is only uglier from the code point of view, but the resulting file is clean.

Nicogene:
> No, the name of the elements is not visible in the cell_arrays.

Ah ok I didn't know about it, I will change the code then, thanks!

traversaro (approved):

Nicogene:
https://github.com/robotology-playground/yarp-telemetry/pull/91/commits/6ecde4dc75282440e1ff1bcecf092500f8928718 should do the trick of the `useless_name`

PR #96, opened 2021-03-12
Add info message when fetching content

S-Dafarra:
CMake hangs while fetching contents. At least we avoid people worrying.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ else()
 
   FetchContent_GetProperties(json)
   if(NOT json_POPULATED)
+    message(STATUS "Fetching nlohmann_json...")
     FetchContent_Populate(json)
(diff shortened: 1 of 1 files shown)


Nicogene (commented):
Does not exist something like `FETCHCONTENT_QUIET` to be set to OFF?
I am not a fetchContent expert

S-Dafarra:
Apparently yes https://cmake.org/cmake/help/latest/module/FetchContent.html:
>FETCHCONTENT_QUIET
The logging output during population can be quite verbose, making the configure stage quite noisy. This cache option (ON by default) hides all population output unless an error is encountered. If experiencing problems with hung downloads, temporarily switching this option off may help diagnose which content population is causing the issue.

Even if I think we are only interested in knowing that something is currently ongoing. Not sure we need all the details

traversaro (approved):

PR #97, opened 2021-03-12
Use push_back instead of emplace_back for the description cell

S-Dafarra:
For some reason, once it crashed here while running a test. In any case, since we want to copy, it does not make any sense to use emplace_back as that will call the copy constructor.

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -250,7 +250,7 @@ public:
         std::scoped_lock<std::mutex> lock{ m_mutex };
         // Add the description
         if (m_description_cell_array.isValid()) {
-            signalsVect.emplace_back(m_description_cell_array);
+            signalsVect.push_back(m_description_cell_array);
         }
         // we have to force the flush.
         flush_all = flush_all || (m_bufferConfig.data_threshold > m_bufferConfig.n_samples);


traversaro:
@S-Dafarra can you provide more details on the failure you are seeing?

S-Dafarra:
Actually, I was not able to reproduce the issue, so probably the fix was not really the one to fix the problem.

By looking at the code, I think it might have been a racing condition on the ``m_description_cell_array`` variable. In particular, it is possible that  while calling the copy constructor on this variable in https://github.com/robotology-playground/yarp-telemetry/blob/ea3bd7ca67c6d72082c19ace59d3b7cad386d7e5/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h#L252-L254
 this may also have been edited by https://github.com/robotology-playground/yarp-telemetry/blob/ea3bd7ca67c6d72082c19ace59d3b7cad386d7e5/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h#L378-L387

Indeed, in https://github.com/robotology-playground/yarp-telemetry/blob/ea3bd7ca67c6d72082c19ace59d3b7cad386d7e5/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h#L386
the internal ``matio`` object gets completely replaced, so this might explain the stack trace I was getting.

So, I suppose that the ``saveToFile`` method was called parallel to ``populateDescriptionCellArray`` and this fix is useless. 

As a side note, in
https://github.com/robotology-playground/yarp-telemetry/blob/ea3bd7ca67c6d72082c19ace59d3b7cad386d7e5/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h#L252 ``isValid`` will always return true, as it has been constructed as a ``CellArray`` (``isValid`` returns false when using the default constructor of ``Variable``, since we don't know the type, or when reading to a file fails).

Nicogene:
> isValid will always return true, as it has been constructed as a CellArray (isValid returns false when using the default constructor of Variable, since we don't know the type, or when reading to a file fails).

Probably it is better a check on the size?

S-Dafarra:
> > isValid will always return true, as it has been constructed as a CellArray (isValid returns false when using the default constructor of Variable, since we don't know the type, or when reading to a file fails).
> 
> Probably it is better a check on the size?

Yes, checking the number of elements (https://dic-iit.github.io/matio-cpp/doxygen/doc/html/classmatioCpp_1_1CellArray.html#a49536bc3e506904a5d354e9afbcc9960) might work.

PR #104, opened 2021-03-22
BufferManager: introduce the usage of the condition variable

### LINKED ISSUE #94: When save_periodically is enabled th BM blocks the executable on closure
Since the save thread is sleeping between one save and the other, when you close the executable it awaits that it wakes up, this may be annoying if using a long period(e.g. 2 minutes).

We may change the handle of the thread with a signaling system(e.g. [conditional variables](https://en.cppreference.com/w/cpp/thread/condition_variable)) in order to wake it up and close.

Nicogene:
This is used for handling the save thread and this should avoid awaiting
the thread save_period seconds to wake up in the closure phase.

It fixes #94

cc @S-Dafarra @lornat75

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
@@ -25,6 +25,7 @@
 #include <thread>
 #include <atomic>
 #include <mutex>
+#include <condition_variable>
 
 
 namespace yarp::telemetry {
@@ -66,8 +67,13 @@ public:
      *
      */
     ~BufferManager() {
-        m_should_stop_thread = true;
         if (m_save_thread.joinable()) {
+            // This additional brackets are needed for make unique_lock out of scope before the join
+            {
+                std::unique_lock<std::mutex> lk_cv(m_mutex_cv);
+                m_should_stop_thread = true;
+                m_cv.notify_one(); // Wake up the thread in order to close it
+            }
             m_save_thread.join();
         }
         if (m_bufferConfig.auto_save) {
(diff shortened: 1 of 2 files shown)


Nicogene:
Seems that with this change the fail on macOs happens more frequently, but I expect that it will be fixed after #9.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
```diff
@@ -356,18 +362,17 @@ class BufferManager {
     }
     void periodicSave()
     {
-        while (!m_should_stop_thread)
+        std::unique_lock<std::mutex> lk_cv(m_mutex_cv);
+        
+        auto timeout = std::chrono::milliseconds(static_cast<uint32_t>(1000 * m_bufferConfig.save_period));
```
traversaro, line 367: Just a curiosity, why not using `std::chrono::seconds`?

traversaro (approved):
Minor comment.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
Nicogene: Because I am afraid that for having the fractions of seconds we should decompose it in seconds/milliseconds.
I am afraid that for example 0.010 will be cast to 0 seconds, using directly milliseconds would workaround this problem.

But I am not a super expert of `std::chrono`

traversaro: Ahh, I see.

Nicogene:
https://github.com/robotology-playground/yarp-telemetry/pull/104/commits/e8e50965680fe88037cbd6415816b9ffb19395dc should test this improvement. With the previous implementation, the execution would remain stuck at `save_thread.join()` for one hour before closing.

src/libYARP_telemetry/src/yarp/telemetry/BufferManager.h
drdanz: You should be able to use `std::chrono::duration<double>(m_bufferConfig.save_period)`

Issue #101, opened 2021-03-15
telemetryDeviceDumper: add documentation


Nicogene:
We should add Doxygen documentation with a table of the parameters and some example of configuration files

Nicogene:
#106 partially fixes this, it add the doc in the readme, the Doxygen part is still missing.

traversaro:
> 
> 
> #106 partially fixes this, it add the doc in the readme, the Doxygen part is still missing.

I think that the README is what most users will look at to use the device, so I think we can close this issue, and eventually we can open a new one to track further improvements.

Nicogene:
Fixed by #106

PR #106, opened 2021-03-24
TelemetryDeviceDumper: make the BM configurable from xml file

### LINKED ISSUE #100: telemetryDeviceDumper: support configuration for the BM
The configuration of the BufferManager is for now hardcoded. We should add the possibility to set it from JSON file of even specifying the single parameters

Nicogene:
It fixes #100

Before proceeding with #101, I would settle which parameters are configurable from the file and which ones are mandatory.

In this implementation, the configuration of the BM from JSON will overwrite the parameter-by-parameters configuration.
`n_samples` is mandatory, as it is mandatory also `save_period` if `save_periodically` is set to true.
At the end it checks, if both `save_periodically` and `auto_save` are set to false the configuration will fail and the device will close.

An issue that I see is that the configuration with the JSON and the configuration parameter-by-parameter are different.
`experiment_name` is `file_name` in the JSON configuration(this probably has to be fixed) and in the JSON configuration ALL the parameters are mandatory for https://github.com/robotology-playground/yarp-telemetry/issues/39#issuecomment-774152259 (also this has to be investigated)

THIS HAS TO BE TESTED ON THE ROBOT.

cc @S-Dafarra

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 ![YARP logo](https://raw.githubusercontent.com/robotology/yarp/master/doc/images/yarp-robot-24.png "YARP")
 YARP telemetry
 ==============
+
 :warning: LIBRARY UNDER DEVELOPMENT :warning:
 
 ![Continuous Integration](https://github.com/robotology-playground/yarp-telemetry/workflows/Continuous%20Integration/badge.svg)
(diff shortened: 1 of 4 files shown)


traversaro (approved):
No need to do it know, but I think the sooner we add the README for the telemetryDeviceDumper, the better, as we can start incrementally documenting any new parameter that we add (even by just forwarding to the library docs), instead of accumulating docs deps.

Nicogene:
https://github.com/robotology-playground/yarp-telemetry/pull/106/commits/d6b79d508814908b8d6690b96203f8885af50b5c adds `telemetryDeviceDumper` to the README

cc @traversaro @AlexAntn @S-Dafarra

traversaro (approved):

Nicogene:
I would merge it and then test on the robot, if there is a bug in the configuration we can open another PR since we have not yet a stable version to maintain

PR #108, opened 2021-03-25
telemetryDeviceDumper: fix xml files and add xml files for running with gazebo

Nicogene:
Today @AlexAntn  spotted that `save_period 120` make fail the configuration since it is not a float(120.0 is fine, we may relax it)

I moreover added the XML and in files for running the `tdd` with gazebo.

Just:
```
$ gazebo
$ cd yarp-telemetry/src/telemetryDeviceDumper/app
$ yarprobotinterface
```

(We can then think where to install those files)


AlexAntn (approved):

traversaro:
> Today @AlexAntn spotted that `save_period 120` make fail the configuration since it is not a float(120.0 is fine, we may relax it)

Shouldn't we simply add a isInt or similar in https://github.com/robotology-playground/yarp-telemetry/blob/master/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp#L133 ?

Nicogene:
> Shouldn't we simply add a isInt or similar in https://github.com/robotology-playground/yarp-telemetry/blob/master/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp#L133 ?

We have to check if isInt fails if you put 0.3 for example  (cc @drdanz ) but yes we could check both eventually

traversaro:
> 
> 
> > Shouldn't we simply add a isInt or similar in https://github.com/robotology-playground/yarp-telemetry/blob/master/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp#L133 ?
> 
> We have to check if isInt fails if you put 0.3 for example (cc @drdanz ) but yes we could check both eventually

I think that indeed isInt will fail for 0.3. but I guess we should accept both int and double, right? If isInt is true do asInt, if isFloating is true do asFloat etc etc.

PR #125, opened 2021-04-20
telemetryDeviceDumper: added all the controlBoardDumper interfaces

### LINKED ISSUE #113: Permit to log data in telemetryDeviceDumper exposed by other controlBoardInterfaces
For backward compatibility, it would be great to match what we support in https://robotology.github.io/robotology-documentation/doc/html/group__icub__controlBoardDumper.html .

Nicogene:
It fixes #113
There are still open points, for now it simply dump everythin, we should define how to
group these quantities under options.

TODO

- [x] Define proper options
- [x] Think about the possibility of using more than one BufferManager
- [x] Think if it is better to just convert CM and IM to doubles
- [x] Write documentation
- [x] Test
- [ ] BONUS: get rid of copy-paste code

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
 | `logJointVelocity`  (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint velocities.     |
 | `logJointAcceleration` (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
 | `logIEncoders`     | bool     | -     |  true | No     | Enable the log of `encoders`, `velocity` and `acceleration` (http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html)     |
-| `logIToequeControl`     | bool     | -     | false     | No     | Enable the log of `torque`(http://yarp.it/git-master/classyarp_1_1dev_1_1ITorqueControl.html).     |
+| `logITorqueControl`     | bool     | -     | false     | No     | Enable the log of `torque`(http://yarp.it/git-master/classyarp_1_1dev_1_1ITorqueControl.html).     |
 | `logIMotorEncoders`     | bool     | -     | false     | No     | Enable the log of `motor_encoders`, `motor_velocity` and `motor_acceleration` (http://yarp.it/git-master/classyarp_1_1dev_1_1IMotorEncoders.html).     |
 | `logIControlMode`     | bool     | -     | false     | No     | Enable the log of `control_mode` (http://yarp.it/git-master/classyarp_1_1dev_1_1IControlMode.html.     |
 | `logIInteractionlMode`     | bool     | -     | false     | No     | Enable the log of `interaction_modes` (http://yarp.it/git-master/classyarp_1_1dev_1_1IInteractionMode.html.     |
@@ -300,7 +300,22 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
 | `data_threshold`     | size_t     | -     | 0     | No     | The save thread saves to a file if there are at least `data_threshold` samples     |
 | `auto_save`     | bool     | -     | false     | No(but it has to be set to true if `save_periodically` is set to false)     | the flag for enabling the save in the destructor of the `yarp::telemetry::experimental::BufferManager`     |
 
+### Mapping .mat variables -> YARP interfaces
 
+| Variable name        | YARP interface |
+| -------------------- | -------------- |
+| `encoders`           | http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html#abcfe10041280b99c7c4384c4fd93a9dd |
+| `velocity`           | http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html#ac84ae2f65f4a93b66827a3a424d1f743 |
+| `acceleration`       | http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html#a3bcb5fe5c6a5e15e57f4723bbe12c57a |
+| `motor_encoders`     | http://yarp.it/git-master/classyarp_1_1dev_1_1IMotorEncoders.html#ac02fb05bb3e9ac9a381d41b59c38c412 |
+| `motor_velocity`     | http://yarp.it/git-master/classyarp_1_1dev_1_1IMotorEncoders.html#a4624348cc129bfeb12ad0e6d2892b76c |
+| `motor_acceleration` | http://yarp.it/git-master/classyarp_1_1dev_1_1IMotorEncoders.html#a9394d8b5cc4f3d58aeaa07c3fb9a6e6a |
+| `control_mode`       | http://yarp.it/git-master/classyarp_1_1dev_1_1IControlMode.html#a32f04715873a8099ec40671f65faff8d |
+| `interaction_mode`   | http://yarp.it/git-master/classyarp_1_1dev_1_1IInteractionMode.html#a6055ce20216f479da6c63807a4d11f54 |
+| `position_error`     | http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#aea29e0fdf34f819ac69a3b940556ba28 |
+| `position_reference` | http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#a29e8f684a15d859229a9ae2902f886da |
+| `torque_error`       | http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#aea29e0fdf34f819ac69a3b940556ba28 |
+| `torque_reference`   | http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#a29e8f684a15d859229a9ae2902f886da |
 
 ### Example of xml
 
--- a/README.md
+++ b/README.md
@@ -277,12 +277,19 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
 ### Parameters
 
 
-
 | Parameter name | Type | Units | Default | Required | Description |
 | -------- | -------- | -------- | -------- | -------- | -------- |
 | `axesNames`     | List of strings     | -  | -     | Yes     | The axes contained in the axesNames parameter are then mapped to the wrapped controlboard in the attachAll method, using controlBoardRemapper class. |
-| `logJointVelocity`     | bool     | -     | false     | No     | Enable the log of joint velocities.     |
-| `logJointAcceleration`     | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
+| `logJointVelocity`  (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint velocities.     |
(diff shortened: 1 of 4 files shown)


Nicogene:
| Parameter name | Type | Units | Default | Required | Description |
| -------- | -------- | -------- | -------- | -------- | -------- |
| `axesNames`     | List of strings     | -  | -     | Yes     | The axes contained in the axesNames parameter are then mapped to the wrapped controlboard in the attachAll method, using controlBoardRemapper class. |
| `logJointVelocity` (DEPRECATED)     | bool     | -     | false     | No     | Enable the log of joint velocities.     |
| `logJointAcceleration` (DEPRECATED)     | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
| `logJointAcceleration` (DEPRECATED)     | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
| `logJointAcceleration` (DEPRECATED)     | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
| `saveBufferManagerConfiguration`     | bool     | -    | false     | No     | Enable the save of the configuration of the BufferManager into `path`+ `"bufferConfig"` + `experimentName` + `".json"`     |
| `json_file`     | string     | -     | -     | No     | Configure the `yarp::telemetry::experimental::BufferManager`s reading from a json file like [in Example configuration file](#example-configuration-file). Note that this configuration will overwrite the parameter-by-parameter configuration   |
| `experimentName`     | string     | -     | -     | Yes     | Prefix of the files that will be saved. The files will be named: `experimentName`+`timestamp`+ `".mat"`.     |
| `path`     | string     | -     | -     | No     | Path of the folder where the data will be saved.     |
| `n_samples`     | size_t     | -     | -     | Yes     | The max number of samples contained in the circular buffer/s     |
| `save_periodically`     | bool     | -     | false     | No(but it has to be set to true if `auto_save` is set to false)     | The flag for enabling the periodic save thread.     |
| `save_period`     | double     | seconds     | -     | Yes(if `save_periodically` is set to true)     | The period in seconds of the save thread     |
| `data_threshold`     | size_t     | -     | 0     | No     | The save thread saves to a file if there are at least `data_threshold` samples     |
| `auto_save`     | bool     | -     | false     | No(but it has to be set to true if `save_periodically` is set to false)     | the flag for enabling the save in the destructor of the `yarp::telemetry::experimental::BufferManager`     |

Nicogene:
Today @AlexAntn and I added a new set of options for managing the new quantities dumped by the `telemetryDeviceDumper`, we deprecated `logJointVelocity/Acceleration` in favor of `logEncoderQuantites` that covers both.

We introduce a list of options that you can see in the following table.
Once we agree on those options we move that table to the README.

| Parameter name | Type | Units | Default | Required | Description |
| -------- | -------- | -------- | -------- | -------- | -------- |
| `axesNames`     | List of strings     | -  | -     | Yes     | The axes contained in the axesNames parameter are then mapped to the wrapped controlboard in the attachAll method, using controlBoardRemapper class. |
| `logJointVelocity`  (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint velocities.     |
| `logJointAcceleration` (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
| `logEncoderQuantities`     | bool     | -     |  true | No     | Enable the log of encoders quantities(http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html)     |
| `logTorqueQuantities`     | bool     | -     | false     | No     | Enable the log of torque quantities(http://yarp.it/git-master/classyarp_1_1dev_1_1ITorqueControl.html).     |
| `logMotorQuantities`     | bool     | -     | false     | No     | Enable the log motor quantities(http://yarp.it/git-master/classyarp_1_1dev_1_1IMotorEncoders.html).     |
| `logModesQuantities`     | bool     | -     | false     | No     | Enable the log of modes quantities(both http://yarp.it/git-master/classyarp_1_1dev_1_1IControlMode.html and http://yarp.it/git-master/classyarp_1_1dev_1_1IInteractionMode.html.     |
| `logPIDQuantities`     | bool     | -     | false     | No     | Enable the log of PID quantities(http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html).|
| `logAmplifierQuantities`     | bool     | -     | false     | No     | Enable the log of amplifier quantities(http://yarp.it/git-master/classyarp_1_1dev_1_1IAmplifierControl.html).     |
| `logAllQuantities`     | bool     | -     | false     | No     | Enable the log all quantities described above. |
| `saveBufferManagerConfiguration`     | bool     | -    | false     | No     | Enable the save of the configuration of the BufferManager into `path`+ `"bufferConfig"` + `experimentName` + `".json"`     |
| `json_file`     | string     | -     | -     | No     | Configure the `yarp::telemetry::experimental::BufferManager`s reading from a json file like [in Example configuration file](#example-configuration-file). Note that this configuration will overwrite the parameter-by-parameter configuration   |
| `experimentName`     | string     | -     | -     | Yes     | Prefix of the files that will be saved. The files will be named: `experimentName`+`timestamp`+ `".mat"`.     |
| `path`     | string     | -     | -     | No     | Path of the folder where the data will be saved.     |
| `n_samples`     | size_t     | -     | -     | Yes     | The max number of samples contained in the circular buffer/s     |
| `save_periodically`     | bool     | -     | false     | No(but it has to be set to true if `auto_save` is set to false)     | The flag for enabling the periodic save thread.     |
| `save_period`     | double     | seconds     | -     | Yes(if `save_periodically` is set to true)     | The period in seconds of the save thread     |
| `data_threshold`     | size_t     | -     | 0     | No     | The save thread saves to a file if there are at least `data_threshold` samples     |
| `auto_save`     | bool     | -     | false     | No(but it has to be set to true if `save_periodically` is set to false)     | the flag for enabling the save in the destructor of the `yarp::telemetry::experimental::BufferManager`     |

cc @S-Dafarra @traversaro

traversaro:
Ok for me, I just suggest to have a more direct connection between names of the C++ interfaces and the options, like the option related to `IEncoders` is `logEncoderQuantities`, but the one related to `ITorqueControl` is `logTorqueQuantities`, so the connection is not obvious, perhaps options like `log<InterfaceName>` lile `logIEncoders`  or  `logITorqueControl` would be less ambigous? 

Furthermore, I think we also need to specify for each option with quantities we log. An issue that unfortunately it is not clear from the YARP interface level and that is instead implementation specific is that while some quantities are already available in physical device such [`embObjMotionControl`](https://github.com/robotology/icub-main/tree/master/src/libraries/icubmod/embObjMotionControl), so calling the related method is not expensive is just copying some buffers, while some other methods are blocking and generate traffic at the firmware level to retrieve online the quantities from the firmware, so having a lot of those calls could be problematic (as far as I know which quantities are cached in `embObjMotionControl` and which one are asked on the fly to the firmware is not documented, but it is probably better to ask to @marcoaccame) . I guess we can discuss in person about that, but in general testing the device on the real iCub robot head with all options enabled would probably be sufficient to validate we are not doing anything wrong.

traversaro:
> We miss the test on iCub in order to check if all the logs are fine (I expect that in gazebo we miss most of them).

Yes, furthermore on Gazebo you do not have the problem of some quantities that are read via expensive RPC calls to the firmware.

README.md
```diff
@@ -277,12 +277,19 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
...
-
 | Parameter name | Type | Units | Default | Required | Description |
 | -------- | -------- | -------- | -------- | -------- | -------- |
 | `axesNames`     | List of strings     | -  | -     | Yes     | The axes contained in the axesNames parameter are then mapped to the wrapped controlboard in the attachAll method, using controlBoardRemapper class. |
-| `logJointVelocity`     | bool     | -     | false     | No     | Enable the log of joint velocities.     |
-| `logJointAcceleration`     | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
+| `logJointVelocity`  (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint velocities.     |
+| `logJointAcceleration` (DEPRECATED)   | bool     | -     | false     | No     | Enable the log of joint accelerations.     |
+| `logIEncoders`     | bool     | -     |  true | No     | Enable the log of `encoders`, `velocity` and `acceleration` (http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html)     |
+| `logIToequeControl`     | bool     | -     | false     | No     | Enable the log of `torque`(http://yarp.it/git-master/classyarp_1_1dev_1_1ITorqueControl.html).     |
```
S-Dafarra, line 286:
```suggestion
| `logITorqueControl`     | bool     | -     | false     | No     | Enable the log of `torque`(http://yarp.it/git-master/classyarp_1_1dev_1_1ITorqueControl.html).     |
```

src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp
```diff
@@ -264,21 +327,66 @@ void TelemetryDeviceDumper::resizeBuffers(int size) {
...
 
-    if (ok && settings.logJointAcceleration) {
+    if (ok && (settings.logJointAcceleration || settings.logIEncoders || settings.logAllQuantities)) {
         ok = ok && bufferManager.addChannel({ "acceleration", {jointAcc.size(), 1} });
     }
 
-    ok = ok && bufferManager.configure(m_bufferConfig);
+    // TODO check if it is more convenient having more BM
+    if (ok && (settings.logIPidControl || settings.logAllQuantities)) {
+        ok = ok && bufferManager.addChannel({ "position_error", {jointPosErr.size(), 1} });
```
traversaro, line 363: It would be nice  to have a table that describes this names, for which options are enabled, and to which interface they are related (nothing too complex, just a table somewhere).

Nicogene: Do you mean something like this https://github.com/robotology/yarp-telemetry/pull/125/commits/78b90ad85c2da977a87b0b81e2dab2921f144b44 ?

README.md
```diff
@@ -293,7 +300,22 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
 | `data_threshold`     | size_t     | -     | 0     | No     | The save thread saves to a file if there are at least `data_threshold` samples     |
 | `auto_save`     | bool     | -     | false     | No(but it has to be set to true if `save_periodically` is set to false)     | the flag for enabling the save in the destructor of the `yarp::telemetry::experimental::BufferManager`     |
 
-
+### Mapping .mat variables -> YARP interfaces
+
+| Variable name        | YARP interface |
+| -------------------- | -------------- |
+| `encoders`           | http://yarp.it/git-master/classyarp_1_1dev_1_1IEncoders.html#abcfe10041280b99c7c4384c4fd93a9dd |
```
traversaro, line 307: This table is not easily readable, perhaps we can change it like this to make it clear which methods are called:
```suggestion
| `encoders`           | [`yarp::dev::IEncoders::getEncoders`](http://yarp.it/latest/classyarp_1_1dev_1_1IEncoders.html#abcfe10041280b99c7c4384c4fd93a9dd) |

Furthermore, I think it is safer to use the latest release of YARP, rather then the git master branch (that is unstable and can contain changes that still needs to be released).
```

Nicogene:
Here is how the mapping table looks like after @traversaro suggestion 

immagine

I think that this PR is ready to be merged

traversaro (approved):

traversaro:
Ok for me!

Issue #128, opened 2021-04-23
YARP_telemetry target should depend on YARP::YARP_conf target


drdanz:
On R1 we had this error:
```
<yarp-telemetry>/src/yarp/telemetry/experimental/api.h:13:10: fatal error: yarp/conf/api.h: No such file or directory
```

Issue #127, opened 2021-04-23
yarp-telemetry does not generate a plugin.ini 


drdanz:
Note: This might actually be an issue in YARP... I'll start the discussion here, and then fix it in YARP.

yarp-telemetry does not call `yarp_configure_plugins_installation` and uses instead YARP directories to install the plugins, therefore the plugin `.ini`/manifest file is not generated.
This is perfectly fine for the install tree version (assuming that it is installed in the same prefix as YARP), but does not work for the build tree version and if it is installed in a different prefix.

I think that there should be a way to generate the .ini for the build directory (so that it can be used with `YARP_DATA_DIRS`), that is not installed if the installation prefix is the same (or perhaps not installed at all). Perhaps this makes sense only if you choose to use YARP installation directories. What do you think?

drdanz:
one option could be make a special case if `YARP_CONFIGURE_PLUGINS_INSTALLATION` is called with `yarp` as component

traversaro:
If I remember well, `yarp_configure_plugins_installation` was indeed introduced to support the cases in which the repo was installing stuff in YARP directories without using `yarp_configure_external_installation` (see https://github.com/robotology/yarp/blob/5313b2defd8be7237567ea56eca8dde93dfc8121/doc/release/v2_3_70.md#cmake-modules-1 and https://github.com/robotology/yarp/pull/1111). So perhaps the solution is just to add a call to `yarp_configure_plugins_installation` ?

drdanz:
If you want to use a different name (for example `yarp_configure_plugins_installation(yarp-telemetry)`) that should just work, but if you use `yarp_configure_plugins_installation(yarp)` it will overwrite the original `yarp.ini` which is not a good behaviour. That's why I was suggesting a special case when you specify `yarp` as component

traversaro:
I think we can just use `yarp_configure_plugins_installation(yarp-telemetry)`, it would work fine without complexity both if YARP is installed in the same prefix or not.

PR #130, opened 2021-04-23
Fix ini generation and compilation

Nicogene:
This PR fixes some issues( #127 and #128)

In particular for #127 this is what I got

```
Directory: C:\Users\ngenesio\robotology\yarp-telemetry\build\x64-Debug\share\yarp\plugins


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         4/23/2021   1:31 PM            164 telemetryDeviceDumper.ini
-a----         4/23/2021   1:31 PM            190 yarp-telemetry.ini
```

The latter one contains:
```
###### This file is automatically generated by CMake.
[search yarp-telemetry]
path "C:/Users/ngenesio/robotology/yarp-telemetry/build/x64-Debug/lib/yarp"
extension ".dll"
type "shared"

```

Is it correct?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,7 +5,9 @@
 # BSD-3-Clause license. See the accompanying LICENSE file for details.
 
 add_subdirectory(libYARP_telemetry)
-add_subdirectory(telemetryDeviceDumper)
-if(BUILD_EXAMPLES)
+if(YARP_os_FOUND AND YARP_dev_FOUND)
+  add_subdirectory(telemetryDeviceDumper)
+endif()
+if(BUILD_EXAMPLES AND YARP_os_FOUND)
   add_subdirectory(examples)
 endif()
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -5,5 +5,7 @@
 # BSD-3-Clause license. See the accompanying LICENSE file for details.
 
 add_subdirectory(libYARP_telemetry)
-add_subdirectory(examples)
 add_subdirectory(telemetryDeviceDumper)
+if(BUILD_EXAMPLES)
+  add_subdirectory(examples)
+endif()
```
traversaro, line 11: New line. : )

drdanz:
The only annoying thing about calling find_package(YARP) more than once is that the "Found YARP" line will be printed multiple times.
I suggest calling it once... something like
```cmake
find_package(YARP ${YARP_REQUIRED_VERSION} COMPONENTS conf OPTIONAL_COMPONENTS os sig dev REQUIRED)
```

you should then pay attention to check
```cmake
if(YARP_os_FOUND)
```
etc

traversaro (approved):

PR #131, opened 2021-04-28
Added possibility to explicitly set the timestamp of a Record with push_back

S-Dafarra:
After a discussion with @prashanthr05, we realized that sometimes it is important to store data in different channels but with the same timestamp. 

In principle, this would be possible by storing structs of data for example, but this is not currently supported. An easy workaround is to store all the data in separated vectors. On the other hand, to make sure that all the data is saved in a synchronized way, it is necessary to specify also the timestamp.

This may also be helpful when storing the data read from the yarp interfaces, saving the timestamp retrieved by the ``yarp`` interface, rather than the one obtained by the application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added the possibility to specify the time of a ``Record`` with ``push_back``.
+
 ## [0.1.0] - 2021-04-02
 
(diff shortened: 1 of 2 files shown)


Nicogene (approved):
👍🏻

S-Dafarra:
I think I did a mistake when stashing and updating. Fixing it

S-Dafarra:
Sorry for the multiple force-pushes:
- First I had applied the modifications copying from an old version of the code. Luckily there where some asserts that did not compile, since the lock of the mutex was not there. Fixed that
- I then removed an assert on the ``now`` function, since it is not necessary in this case.

PR #134, opened 2021-05-17
Fix compilation when using Ubuntu 18.04 

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/731 . 

As discussed in https://github.com/robotology/robotology-superbuild/issues/481, we plan to support the compilation of this library on Ubuntu 18.04 at least in the next future. For this reason, this PR fixes the library to compile correctly when on Ubuntu 18.04, that was broken after https://github.com/robotology/yarp-telemetry/pull/132 .

Please squash when merging the PR, thanks!

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/experimental/BufferManager.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/experimental/BufferManager.h
@@ -26,7 +26,18 @@
 #include <atomic>
 #include <mutex>
 #include <condition_variable>
-#include <filesystem>
+
+#ifndef __has_include
+  static_assert(false, "__has_include not supported");
+#else
+#  if __has_include(<filesystem>)
+#    include <filesystem>
+     namespace yarp_telemetry_fs = std::filesystem;
+#  elif __has_include(<experimental/filesystem>)
+#    include <experimental/filesystem>
+     namespace yarp_telemetry_fs = std::experimental::filesystem;
+#  endif
+#endif
 
 
 namespace yarp::telemetry::experimental {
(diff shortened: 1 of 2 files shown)


src/libYARP_telemetry/src/yarp/telemetry/experimental/BufferManager.h
```diff
@@ -26,7 +26,18 @@
...
+  static_assert(false, "__has_include not supported");
+#else
+#  if __has_include(<filesystem>)
+#    include <filesystem>
+     namespace yarp_telemetry_fs = std::filesystem;
+#  elif __has_include(<experimental/filesystem>)
+#    include <experimental/filesystem>
+     namespace yarp_telemetry_fs = std::experimental::filesystem;
+#  endif
+#endif
```
drdanz, line 40: I have a weak point for checking this using `CheckIncludeFileCXX` instead of `__has_include`, since you could catch an unsupported system at configure time... Anyway probably not worth wasting time on this, since I'm not sure if we support any compiler which doesn't have `__has_include` or `filesystem`/`experimental/filesystem`. It might be worth adding an error if neither of them is available, though.

traversaro: Added error message in https://github.com/robotology/yarp-telemetry/pull/134/commits/17658b990c239e622f537bb0cd8fa1755d2bbdca .  As you said, it is probably not useful to use `CheckIncludeFileCXX`, and I must add that any kind of that checks is quite slow on Windows.

drdanz (approved):

Nicogene (approved):

traversaro:
The macOS/Homebrew jobs are taking more then one hour just to install the dependencies. As macOS/Conda is working fine, what do you think about merging as compilation on macOS is working fine? @Nicogene

traversaro:
> 
> 
> The macOS/Homebrew jobs are taking more then one hour just to install the dependencies. As macOS/Conda is working fine, what do you think about merging as compilation on macOS is working fine? @Nicogene

In the end, the build was successful also with macOS/Homebrew jobs.

Nicogene:
Sorry I forgot to merge, thanks!

PR #138, opened 2021-06-08
telemetryDeviceDumper: add log of Odometry

### LINKED ISSUE #126: Add support for odometry to telemetryDeviceDumper
According to @randaz81 the data he would like to dump on the R1 robot is the odometry data coming from the `ILocalization2D` interface.

Data is generated by the `odomLocalizer` device, and is published on the yarp network by `localization2DServer` (now `localization2D_nws_yarp`)

Nicogene:
This PR adds the possibility to log the odometry data.
It fixes #126 

There are still some open points:

- Differently from the other data retrieved through the attaching and the control_board_remapper, for the odometry we used the `localization2DClient`, since the localization device can run on different machines from the `r1-base`, and the localization pipeline has never been used with `yarprobotinterface`. We have to discuss the pro and cons of following this approach or the "attach" approach
- Now we added the log odometry to `logAllQuantities` option, but it may break some existing configurations, since on icub the open of the `localization2DClient` will fail. Maybe we should divide in control board quantities from the rest?
- Until now we always assumed that the remapper should be opened, otherwise no data would be logged, but maybe we should allow logging just the odometry(or other future quantities) that does not need the remapper?

We tested with [this self-contained example](https://github.com/robotology/navigation/blob/master/app/robotPathPlannerExamples/scripts/robotPathPlannerExample2.xml), that started correctly

immagine

But we get only:
```
[WARNING] telemetryDeviceDumper warning : odometry_data was not readed correctly
[ERROR] |yarp.device.localization2DClient| getEstimatedOdometry is not yet implemented
[WARNING] telemetryDeviceDumper warning : odometry_data was not readed correctly
[ERROR] |yarp.device.localization2DClient| getEstimatedOdometry is not yet implemented
[WARNING] telemetryDeviceDumper warning : odometry_data was not readed correctly
[ERROR] |yarp.device.localization2DClient| getEstimatedOdometry is not yet implemented
[WARNING] telemetryDeviceDumper warning : odometry_data was not readed correctly
[ERROR] |yarp.device.localization2DClient| getEstimatedOdometry is not yet implemented
``` 

In any case on Friday @AlexAntn should do some tests with r1

TODO: update the documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added the log of the estimated odometry from `yarp::dev::Nav2D::ILocalization2D`.
+- Deprecated the `logAllQuantities` option in favour of `logControlBoardQuantities`.
+
 ## [0.2.0] - 2021-05-19
 
 - Added the possibility to specify the time of a ``Record`` with ``push_back``.
--- a/README.md
+++ b/README.md
@@ -294,7 +294,9 @@ The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_de
 | `logIInteractionMode`     | bool     | -     | false     | No     | Enable the log of `interaction_modes` (http://yarp.it/git-master/classyarp_1_1dev_1_1IInteractionMode.html.     |
 | `logIPidControl`     | bool     | -     | false     | No     | Enable the log of `position_error`, `position_reference`, `torque_error`, `torque_reference`(http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html).|
 | `logIAmplifierControl`     | bool     | -     | false     | No     | Enable the log of `pwm` and `current` (http://yarp.it/git-master/classyarp_1_1dev_1_1IAmplifierControl.html).     |
-| `logAllQuantities`     | bool     | -     | false     | No     | Enable the log all quantities described above. |
+| `logAllQuantities` (DEPRECATED)    | bool     | -     | false     | No     | Enable the log all quantities described above. |
+| `logControlBoardQuantities` | bool     | -     | false     | No     | Enable the log of all the quantities that requires the attach to a control board (`logIEncoders`, `logITorqueControl`, `logIMotorEncoders`, `logIControlMode`, `logIInteractionMode`, `logIPidControl`, `logIAmplifierControl`). |
+| `logILocalization2D` | bool     | -     | false     | No     | Enable the log of `odometry_data` (http://yarp.it/git-master/classyarp_1_1dev_1_1Nav2D_1_1ILocalization2D.html). For properly logging the odometry data [this patch](https://github.com/robotology/yarp/pull/2602) is needed on YARP side |
 | `saveBufferManagerConfiguration`     | bool     | -    | false     | No     | Enable the save of the configuration of the BufferManager into `path`+ `"bufferConfig"` + `experimentName` + `".json"`     |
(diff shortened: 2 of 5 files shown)


traversaro:
The strange thing is that after this PR, any use of `logAllQuantities` will fail unless there is a a valid odometryserver in the YARP network.

randaz81:
`getEstimatedOdometry is not yet implemented. `  I think it's a regression bug of the odomLocalizer plugin. I'm going to fix it now.

Nicogene:
> The strange thing is that after this PR, any use of logAllQuantities will fail unless there is a a valid odometryserver in the YARP network.

In fact, I would like to keep it outside the `logAllQuantities` option, but this means it is not `all` anymore

randaz81:
> `getEstimatedOdometry is not yet implemented. ` I think it's a regression bug of the odomLocalizer plugin. I'm going to fix it now.

fixed now in navigation master branch

Nicogene:
@randaz81 I still get the same error since in `localization2DClient` `getEstimatedOdometry` is not implemented:

https://github.com/robotology/yarp/blob/9cdf3b716b5ba7e2c8038358bcd2d73503b896f4/src/devices/localization2DClient/Localization2DClient.cpp#L162-L166

It is the right way to get it?

Nicogene:
After https://github.com/robotology/yarp/pull/2602 it worked, and here it is the resulting mat file:

```
test_telemetry = 

  struct with fields:

    description_list: {32×1 cell}
        acceleration: [1×1 struct]
            velocity: [1×1 struct]
       odometry_data: [1×1 struct]
            encoders: [1×1 struct]
```


```
test_telemetry.odometry_data

ans = 

  struct with fields:

          data: [9×1×11815 double]
    dimensions: [9 1 11815]
          name: 'odometry_data'
    timestamps: [1×11815 double]
```

This is the data running  https://github.com/robotology/navigation/blob/master/app/robotPathPlannerExamples/scripts/robotPathPlannerExample2.xml 

immagine

Nicogene:
We have just to understand how to treat the option for enabling it, and the fact that the odometry part for working need a newer version of yarp(just run time issue, not compile time)

cc @traversaro @AlexAntn @vtikha

traversaro:
> We have just to understand how to treat the option for enabling it,

In which sense?

>  and the fact that the odometry part for working need a newer version of yarp(just run time issue, not compile time)

For this, I think docs is enough.

Nicogene:
> In which sense?

In the sense that if we add it to `logAllQuantities` option, or we exclude it and rename the option to `logCBQuantities` for this https://github.com/robotology/yarp-telemetry/pull/138#issuecomment-856888309

traversaro:
> 
> 
> > In which sense?
> 
> In the sense that we have to decide if we add it to `logAllQuantities` option, or we exclude it and rename the option to `logCBQuantities` for this [#138 (comment)](https://github.com/robotology/yarp-telemetry/pull/138#issuecomment-856888309)

I see. My opinion is that either `logAllQuantities` is converted to only log the quantities of the interfaces that are available in the attached devices, or indeed thematic options such as `logControlBoardQuantities` make more sense.

Nicogene:
https://github.com/robotology/yarp-telemetry/pull/138/commits/9a8d1ff2b3ddc9b62a7a46e393b232841dfcd8f8 removes `odometry_data` to `logAllQuantities`

https://github.com/robotology/yarp-telemetry/pull/138/commits/c2a6a7bab7c97f324c7afed0810e1404b869ad25 deprecates `logAllQuantities` in favour of `logControlBoardQuantities`

The documentation and changelog have been updated accordingly

traversaro (approved):

PR #144, opened 2021-12-09
Introduce the `matioCpp::Span` in the interface of yarp-telemetry

### LINKED ISSUE #49: Investigate about a more flexible data type for storing data
From #37:
> Right now the interface requires a reference to a std::vector as input. This is not the most flexible data type when it comes to vectors. In our applications, we use std::vectors just for a narrow subset of data. For all the other cases, we would need an intermediate copy, which is not so convenient. In order to avoid dynamic memory allocation in time-critical methods, we would need to use a std::vector as a class parameter for each variable we want to store, which is not very practical.

In particular, we have to consider that is related to what is under the hood of `yarp::telemetry::Record`
A similar discussion has been done in #23.

GiulioRomualdi:
This closes #49 

this is the first step towards https://github.com/ami-iit/bipedal-locomotion-framework/issues/189

cc @traversaro @S-Dafarra @Nicogene

### DIFF
--- a/src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
@@ -9,6 +9,8 @@
 #ifndef YARP_TELEMETRY_RECORD_H
 #define YARP_TELEMETRY_RECORD_H
 
+#include <vector>
+
 #include <matioCpp/Span.h>
 
 namespace yarp::telemetry::experimental {
@@ -37,6 +39,29 @@ struct Record
      */
     Record(const double& _ts,
            matioCpp::Span<const T> _datum) : m_ts(_ts), m_datum(_datum.begin(), _datum.end()) {
+               m_payload = sizeof(m_ts) + sizeof(m_datum) + sizeof(T) * m_datum.capacity();
+    }
+
+    /**
+     * @brief Construct a new Record object copying the _datum
+     *
+     * @param[in] _ts Timestamp to assign to the record.
+     * @param[in] _datum Datum to be copied.
+     */
+    Record(const double& _ts,
+           const std::initializer_list<T>& _datum)
+        : m_ts(_ts), m_datum(_datum.begin(), _datum.end()) {
+               m_payload = sizeof(m_ts) + sizeof(m_datum) + sizeof(T) * m_datum.capacity();
+    }
+
+    /**
+     * @brief Construct a new Record object moving the _datum
+     *
+     * @param[in] _ts Timestamp to assign to the record.
+     * @param[in] _datum Datum to be moved.
+     */
+    Record(const double& _ts,
+           std::vector<T>&& _datum) : m_ts(_ts), m_datum(std::move(_datum)) {
                m_payload = sizeof(m_ts) + sizeof(m_datum) + sizeof(T) * m_datum.capacity();
     }
 
--- a/src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
+++ b/src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
@@ -9,7 +9,7 @@
 #ifndef YARP_TELEMETRY_RECORD_H
 #define YARP_TELEMETRY_RECORD_H
 
-#include <vector>
+#include <matioCpp/Span.h>
 
 namespace yarp::telemetry::experimental {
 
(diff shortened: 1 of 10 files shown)


AlexAntn (approved):
LGTM!

src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
```diff
@@ -36,20 +36,10 @@ struct Record
      * @param[in] _datum Datum to be copied.
      */
     Record(const double& _ts,
-           const std::vector<T>& _datum) : m_ts(_ts), m_datum(_datum) {
+           matioCpp::Span<const T> _datum) : m_ts(_ts), m_datum(_datum.begin(), _datum.end()) {
```
Nicogene, line 39: My preface is that I didn't check how `matioCpp::Span` is implemented and work, my concern here is the possible API/ABI break.
I think that the ABI break would be inevitable, but also negligible, but this breaks the API?
I mean, usage of the span is transparent to the std::vector?

src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
```diff
@@ -36,20 +36,10 @@ struct Record
...
-    /**
-     * @brief Construct a new Record object moving the _datum
-     *
-     * @param[in] _ts Timestamp to assign to the record.
-     * @param[in] _datum Datum to be moved.
-     */
-    Record(const double& _ts,
-           std::vector<T>&& _datum) : m_ts(_ts), m_datum(std::move(_datum)) {
-               m_payload = sizeof(m_ts) + sizeof(m_datum) + sizeof(T) * m_datum.capacity();
-    }
```
Nicogene, line 52: Why do we need to remove the move operators?

test/BufferManagerTest.cpp
```diff
@@ -92,9 +96,11 @@ TEST_CASE("Buffer Manager Test")
         REQUIRE(bm_v.configure(bufferConfig));
 
         for (int i = 0; i < 10; i++) {
-            bm_v.push_back({ i + 1.0, i + 2.0, i + 3.0, i + 4.0 }, "one");
+            std::vector tmp1 = { i + 1.0, i + 2.0, i + 3.0, i + 4.0 };
```
Nicogene, line 99: Why you have to declare the input vector outside? `matioCpp::Span` does not have the initializer_list constructor?
In this case seems an API break

Nicogene (commented):
@GiulioRomualdi thanks for submitting the PR!
I have few ABI/API breaks that concern me but we can discuss about it how handle them in the painless way

S-Dafarra:
What if we have both the possibility to pass a vector **and** a ``span``? We would probably need some SFINAE hack to avoid ambiguity (i.e. the ``span`` method should not work if you are passing a ``std::vector``), but at least in this way we avoid API breaks, and we don't even need to change all the examples :thinking:

test/BufferManagerTest.cpp
GiulioRomualdi: Hi @Nicogene, unfortunately `span` (even the `std`) cannot be initialized with  `initializer_list`. I can add implement the `initializer_list` constructor so we ensure the backcompatibility. What do you think?

```cpp
// printSpan.cpp

#include <iostream>
#include <span>

void printMe(std::span<const int> container) {
    
    std::cout << "container.size(): " << container.size() << '\n';  // (4)
    for(auto e : container) std::cout << e << ' ';
    std::cout << "\n\n";
}

int main() {
    printMe({1, 2, 3, 4});
}
```

```
main.cpp: In function 'int main()':
main.cpp:14:12: error: could not convert '{1, 2, 3, 4}' from '<brace-enclosed initializer list>' to 'std::span<const int>'
   14 |     printMe({1, 2, 3, 4});
      |     ~~~~~~~^~~~~~~~~~~~~~
      |            |
      |            <brace-enclosed initializer list>
```

GiulioRomualdi:
> What if we have both the possibility to pass a vector **and** a `span`? We would probably need some SFINAE hack to avoid ambiguity (i.e. the `span` method should not work if you are passing a `std::vector`), but at least in this way we avoid API breaks, and we don't even need to change all the examples thinking

Hi @S-Dafarra I answered here: https://github.com/robotology/yarp-telemetry/pull/144#discussion_r768457176

src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
GiulioRomualdi: I can reintroduce it!

src/libYARP_telemetry/src/yarp/telemetry/experimental/Record.h
GiulioRomualdi: I think that by reimplementing https://github.com/robotology/yarp-telemetry/pull/144#discussion_r766783307 and https://github.com/robotology/yarp-telemetry/pull/144#discussion_r768457176 the API are kept stable

test/BufferManagerTest.cpp
Nicogene: If it is not too much complicated I would do it, in this way we can also avoid to update alle the examples and the documentation, thanks!

GiulioRomualdi:
In https://github.com/robotology/yarp-telemetry/pull/144/commits/39e96dce98f6798187efba538443ae023df46619 and https://github.com/robotology/yarp-telemetry/pull/144/commits/8eca2ead72f31bc3fd013cd9f067ab58c92bdbb9 I restored the missing methods and I implemented the `std::initializer_list` methods

https://github.com/robotology/yarp-telemetry/pull/144/commits/a48800132cdac90f508e68018ecad4f2f2833640 reintroduces the back-compatibility for the tests and the examples

Nicogene:
The failure of the CI seems unrelated, I restarted it

GiulioRomualdi:
I should update the [changelog](https://github.com/robotology/yarp-telemetry/blob/master/CHANGELOG.md) before the merge

Nicogene (approved):
If the test passes I think we are good to go 👍🏻

Nicogene:
Merged, thanks!

PR #148, opened 2021-12-13
Add the possibility to use a different file indexing for the mat files

GiulioRomualdi:
With this PR it will be possible to support different methods for file indexing. 

For instance setting
```cpp
config.file_indexing = "%Y_%m_%d_%H_%M_%S";
```

you will get a file named `<file_name>_2021_12_13_19_42_08.mat`

cc @Nicogene @AlexAntn @traversaro @S-Dafarra


Nicogene (changes requested):
Thanks for the PR, it seems a very nice addition!

Could you please update the documentation with an example of usage and add an example that uses this new feature ?
Have you tried it with the json configuration? I don't remember if there is something to change in the parsing of the json, I think not, but I would give a try.

Thanks!

GiulioRomualdi:
Hi @Nicogene, I update the example and the README.

Nicogene (approved):

Nicogene:
Merged, thanks!

PR #159, opened 2022-03-16
Add the possibility to call a callback when the buffer is saved

GiulioRomualdi:
This PR implements the possibility to set a callback that is called every time the `saveToFile` is called. Thanks to this it will be possible to store other kinds of data, e.g. videos, txt files...

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,30 @@ ans =
 
 ```
 
+### Example additional callback
+
+``BufferManager`` can call an additional callback every time the save function is called. The
+following example define a custom callback that saves a dummy `txt` file along with the `mat` saved
+by the telemetry
+```c++
+bool myCallback(const std::string& file_name, const SaveCallbackSaveMethod& method) {
+  std::string file_name_with_extension = file_name + ".txt";
+  std::ofstream my_file(file_name_with_extension.c_str());
+
+  // Write to the file
+  my_file << "Dummy file!";
+
+  // Close the file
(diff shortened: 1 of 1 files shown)


GiulioRomualdi:
cc @traversaro @S-Dafarra @Nicogene @AlexAntn

Nicogene (changes requested):
Hi @GiulioRomualdi sorry for the delay!

It is not super clear to me a possible use case, could you add an example of it?

Maybe adding a test and could you please add it in the readme?

Nicogene:
Hi @GiulioRomualdi, what is the status of this PR?
We will do soon a release and then do some breaking changes that would trigger some conflicts with this PR

GiulioRomualdi:
> Hi @GiulioRomualdi, what is the status of this PR? We will do soon a release and then do some breaking changes that would trigger some conflicts with this PR


Hi @Nicogene, sorry for the delay. I updated the readme and the test I also apply what @S-Dafarra suggested. 

An use case for this PR is to save the videostream from the cameras. Thanks to this I can merge this: https://github.com/ami-iit/bipedal-locomotion-framework/pull/516

S-Dafarra (approved):

Nicogene (approved):
I will add a line in the changelog and I will release the v0.5.1 for including this patch before the renaming

Nicogene:
Merged, thanks!

PR #175, opened 2022-05-10
Avoid using templates in BufferManager.

S-Dafarra:
With this PR I completely removed the use of templates for ``BufferManager`` since the default template was in any case causing issues.

I have also fixed a small discrepancy in the definition of the dimensions of cell and struct arrays.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ ans =
 
   struct with fields:
 
-              data: {3×1 cell}
+              data: {1×3 cell}
         dimensions: [1 3]
     elements_names: {'element_0'}
               name: 'string_channel'
--- a/README.md
+++ b/README.md
@@ -138,7 +138,6 @@ buffer_manager_test.one =
 
 It is possible to save and dump also vector variables.
 Here is the code snippet for dumping in a `.mat` file 3 samples of the 4x1 vector variables `"one"` and `"two"`.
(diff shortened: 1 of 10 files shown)


S-Dafarra:
cc @GiulioRomualdi @traversaro

README.md
```diff
@@ -364,7 +363,7 @@ ans =
 
   struct with fields:
 
-              data: {3×1 cell}
+              data: {1×3 cell}
```
Nicogene, line 366: Why did you change it?

Nicogene (changes requested):
Could you please add [this test](https://github.com/robotology/robometry/commit/fac3227c47aefc60e89442ee75f1365ead6d8057) for checking the issue this PR is addressing?

README.md
S-Dafarra: See https://github.com/robotology/robometry/pull/175/commits/a6309eedc8419607dcb68f870bd87fd46eebe14a, I realized that for cell and struct arrays I was concatenating on the first dimension. This was causing differences with the ``dimensions`` vector.

You can actually see this mismatch in the previous version of the README: https://github.com/robotology/robometry/blob/14310e9a80c75eba987d647f02a91afe9f0c0512/README.md?plain=1#L367-L368

Nicogene: Aaa I see

Nicogene (approved):
LGTM!

Is it ok for you @GiulioRomualdi @traversaro ?

traversaro (approved):

PR #178, opened 2022-07-04
Add units_of_measure parameter

Nicogene:
This PR adds `units_of_measure` parameter, it is handled as string, here is an example:

```c++
        robometry::BufferManager bm;
        robometry::BufferConfig bufferConfig;

        robometry::ChannelInfo var_one{ "one", {1,1}, {}, {"m"} };
        robometry::ChannelInfo var_two{ "two", {1,1}, {}, {"deg"}};
        // First add channels that will be handling empty buffers
        REQUIRE(bm.addChannel(var_one));
        REQUIRE(bm.addChannel(var_two));

        bufferConfig.yarp_robot_name = "robot";
        bufferConfig.description_list = { "Be", "Or not to be" };
        bufferConfig.filename = "buffer_manager_test_unit_measure";
        bufferConfig.n_samples = n_samples;
        REQUIRE(bm.configure(bufferConfig));

        for (int i = 0; i < 3; i++) {
            bm.push_back({ i }, "one");
            std::this_thread::sleep_for(std::chrono::milliseconds(10));
            bm.push_back({ i + 1 }, "two");
        }
        // Check manual save
        REQUIRE(bm.saveToFile());

```

And here is the resulting mat file:

```
buffer_manager_test_unit_measure = 

  struct with fields:

    description_list: {2×1 cell}
     yarp_robot_name: 'robot'
                 two: [1×1 struct]
                 one: [1×1 struct]

>> buffer_manager_test_unit_measure.one

ans = 

  struct with fields:

               data: [1×1×3 int32]
         dimensions: [1 1 3]
     elements_names: {'element_0'}
    unit_of_measure: {'m'}
               name: 'one'
         timestamps: [1657011185.64147 1657011185.65175 1657011185.66186]

>> buffer_manager_test_unit_measure.two

ans = 

  struct with fields:

               data: [1×1×3 int32]
         dimensions: [1 1 3]
     elements_names: {'element_0'}
    unit_of_measure: {'deg'}
               name: 'two'
         timestamps: [1657011185.65173 1657011185.66184 1657011185.67198]
```

cc @GiulioRomualdi


GiulioRomualdi:
I didn't go towards the code but please notice that the same channel may have different units. For instance, the FT readouts contain force and torques, so the first three elements are in $N$ while the last three are in $Nm$.

Nicogene:
> That's cool @Nicogene!
> 
> I didn't go towards the code but please notice that the same channel may have different units. For instance, the FT readouts contain force and torques, so the first three elements are in N while the last three are in Nm.

You are right, I thought 1 channel -> 1 unit but it is not always true, maybe it has to be handled as vector of strings

Nicogene:
@GiulioRomualdi I refactored in order to use a vector of strings instead of a string.
If the user specify a just one unit measure, it is all applied to the element of the data, e.g. for a vector 4x1, if the user just specify `{"m"}`, in the resulting mat file it will have a struct like this:
```
  4×1 cell array

    {'m'}
    {'m'}
    {'m'}
    {'m'}
```
If not specified it will create a vector of empty strings:
```
  4×1 cell array

    {''}
    {''}
    {''}
    {''}
```

This is because `CellArray` cannot have dimension = 0:

https://github.com/ami-iit/matio-cpp/blob/a1feb0a79d378c2a22c3c16314d06074dfa906ad/src/CellArray.cpp#L49-L53

traversaro (approved):
I think it would be cool to refer to some kind of standard way of expressing unit of measurements as string, but I could not find anything wild used, so I think we can leave as it is. However, we may want to write something that the units of measure is just a string.

Nicogene:
I have changed the default value to "n.d" instead of "" because matio was complaining about it:
```
[ERROR][matioCpp::Variable::createVar] The name should not be empty.
[ERROR][matioCpp::CellArray::setElement] The input variable is not valid.
```

S-Dafarra:
> I have changed the default value to "n.d." instead of "" because matio was complaining about it:
> 
> ```
> [ERROR][matioCpp::Variable::createVar] The name should not be empty.
> [ERROR][matioCpp::CellArray::setElement] The input variable is not valid.
> ```

This is a bug. I am trying to fix it in https://github.com/ami-iit/matio-cpp/pull/57, but then probably a new release has to be made.

S-Dafarra:
It is probably late, but I just release a new version with the bugfix, in case you want to restore the version with the empty string: https://github.com/ami-iit/matio-cpp/releases/tag/v0.2.1

Nicogene:
> It is probably late, but I just released a new version with the bugfix, in case you want to restore the version with the empty string: https://github.com/ami-iit/matio-cpp/releases/tag/v0.2.1

Thank you! I think that for now, I would leave "n.d." if it doesn't give problems

PR #182, opened 2022-09-06
Switch README to point to conda-forge package of robometry

traversaro:
After https://github.com/conda-forge/staged-recipes/pull/20219, robometry is officially available on conda-forge without the need to use the additional robotology channel.


Nicogene:
Merged, thanks!

Issue #183, opened 2023-08-21
General CI Failure August 2023


traversaro:
https://github.com/robotology/robometry/actions/runs/5921165075
https://github.com/robotology/robometry/actions/runs/5921143453

Nicogene:
Chatting /w @traversaro it came out this is due to a regression of this PR:

- https://github.com/robotology/ycm/pull/422

To fix it we could either:
- port the test to catch3
- Force the usage of ycm 0.14.0 in the CI

PR #184, opened 2023-09-04
CI: Use mambaforge

traversaro:
To avoid incompatibility between defaults channel and conda-forge channel


Nicogene (approved):
The failure are related to #183, merging

PR #187, opened 2024-01-05
Fix number of samples

DanielePucci:
This should fix the number of samples being saved in the snippet examples. All examples, in fact, claim to save 3 samples while all 'for' loops were from 0 to 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Here is the code snippet for dumping in a `.mat` file 3 samples of the scalar va
         return 1;
     }
 
-    for (int i = 0; i < 10; i++) {
+    for (int i = 0; i < 3; i++) {
         bm.push_back(i , "one");
         std::this_thread::sleep_for(std::chrono::milliseconds(200));
         bm.push_back(i + 1.0, "two");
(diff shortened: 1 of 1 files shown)


traversaro (approved):

S-Dafarra:
Actually, I believe that the problem is not in the number of samples, but a missing detail.
If you specify that you want to save 3 samples, ``robometry`` will save a file with three entries, even if you push 10 times. It will simply save the last three items. This behavior may not properly specified in the README.

Moreover, if you enable the ``autosave`` option, it will create a new file every time you push 3 samples. Hence, when pushing 10, you should expect three files with 3 samples each, and 1 file with the last item (triggering the manual "save").

Also this might not be properly specified in the README.

By forcing the for loops to run fr three elements does not improve the clarity in this sense.

Another issue with the README is that in this line https://github.com/robotology/robometry/blob/29646dc3d54037f7ddc8f173f4a01cbd82ba3c78/README.md#L84
``nsamples`` should be substituted with ``3``.

Nicogene (approved):

Nicogene:
I am merging this and for what @S-Dafarra said I opened a dedicated issue, feel free to fill up the missing informations in the readme

PR #190, opened 2024-09-18
Add to TelemetryDeviceDumper streaming of rawValues from low level

MSECode:
Add to the `telemetryDeviceDumper` a device based on the nwc `rawValuesPublisherClient` thus to save in the MATLAB file generated by this plugin the raw values (and whatever one desired for debugs based on the Interface `iCub::debugLibrary::IRawValuesPublisher`) coming from the boards at the fw level.

**Side Note:** I'll open this PR since I've tested the update done on my bench setup. However, the work, even though at a good point, cannot be considered fully completed. The PR will be kept in draft for now and used for discussion on the updates done and to revise them together

cc: @valegagge @Nicogene @traversaro

### DIFF
--- a/src/telemetryDeviceDumper/CMakeLists.txt
+++ b/src/telemetryDeviceDumper/CMakeLists.txt
@@ -20,12 +20,14 @@ if(ENABLE_telemetryDeviceDumper)
     add_definitions(-D_USE_MATH_DEFINES)
   endif()
 
+  find_package(iCubDev REQUIRED)
   target_sources(yarp_telemetryDeviceDumper PRIVATE TelemetryDeviceDumper.cpp
                                                     TelemetryDeviceDumper.h)
-
+                        
   target_link_libraries(yarp_telemetryDeviceDumper PRIVATE YARP::YARP_os
                                                            YARP::YARP_dev
-                                                           robometry::robometry)
+                                                           robometry::robometry
+                                                           ICUB::iCubDev)
   list(APPEND YARP_${YARP_PLUGIN_MASTER}_PRIVATE_DEPS YARP_os
                                                       YARP_dev
                                                       robometry)
--- a/src/telemetryDeviceDumper/CMakeLists.txt
+++ b/src/telemetryDeviceDumper/CMakeLists.txt
@@ -20,7 +20,7 @@ if(ENABLE_telemetryDeviceDumper)
     add_definitions(-D_USE_MATH_DEFINES)
   endif()
 
-  find_package(iCubDev REQUIRED)
+  find_package(iCubDev 2.7.0 REQUIRED)
   target_sources(yarp_telemetryDeviceDumper PRIVATE TelemetryDeviceDumper.cpp
                                                     TelemetryDeviceDumper.h)
                         
--- a/src/telemetryDeviceDumper/TelemetryDeviceDumper.h
+++ b/src/telemetryDeviceDumper/TelemetryDeviceDumper.h
@@ -19,6 +19,7 @@
 #include <yarp/dev/IControlMode.h>
 #include <yarp/dev/IInteractionMode.h>
 #include <yarp/dev/ILocalization2D.h>
+#include <iCub/IRawValuesPublisher.h>
 #include <yarp/dev/IMotor.h>
 #include <yarp/dev/ITorqueControl.h>
 #include <yarp/dev/PolyDriver.h>
@@ -27,6 +28,7 @@
 #include <robometry/BufferManager.h>
 
 #include <unordered_map>
+#include <map>
 #include <string>
 #include <memory>
 #include <vector>
@@ -47,9 +49,12 @@ struct TelemetryDeviceDumperSettings {
     bool logIPidControl{ false };
     bool logIAmplifierControl{ false };
     bool logILocalization2D{ false };
+    bool logIRawValuesPublisher { false };
     bool useRadians{ false };
     bool saveBufferManagerConfiguration{ false };
     std::string localizationRemoteName{ "" };
+    std::string rawValuesPublisherRemoteName { "" };
+
 };
 /**
  * @brief The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_devices.html)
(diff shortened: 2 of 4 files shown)


S-Dafarra:
To be honest, I believe that the dependency on iCub is not the best here, as it would complexify its usage with other robots. An alternative could be to instead create a similar device in icub-main instead

valegagge:
> To be honest, I believe that the dependency on iCub is not the best here, as it would complexify its usage with other robots. An alternative could be to instead create a similar device in icub-main instead

Hi @S-Dafarra ,
I don't understand your suggestion, sorry! :( The new device belongs to icub-main, so here we add the dependence from it.

valegagge:
Hi @MSECode ,
for me is ok.
I found a small bug related to the use of `logControlBoardQuantities` with `false `value, but it seems not dependent on your changes.
I fixed here. Asap I'll put the code (I need the permission.)

S-Dafarra:
> The new device belongs to icub-main, so here we add the dependence from it.

That's the problem, the new dependency from ``icub-main``, while it should be the other way around. In other words, it should be ``icub-main`` to depend from ``robometry``, and this can be done by moving or copying the ``TelemetryDeviceDumper `` device into ``icub-main``.

Nicogene:
> > The new device belongs to icub-main, so here we add the dependence from it.
> 
> That's the problem, the new dependency from `icub-main`, while it should be the other way around. In other words, it should be `icub-main` to depend from `robometry`, and this can be done by moving or copying the `TelemetryDeviceDumper ` device into `icub-main`.

Note that the dependency from `icub-main` would be added only if `telemetryDeviceDumper` is compiled, and it is disabled by default.

In any case, I am more comfortable to move it eventually in `icub-main`, rather than have yet another logger device around

S-Dafarra:
> In any case, I am more comfortable to move it eventually in `icub-main`, rather than have yet another logger device around

I agree! Your call anyway 👍

Nicogene:
Another possibility is to add a `find_package(iCub QUIET)` and then use a `ifdef` in the `telemetryDeviceDumper` code, this would keep it compilable also without `icub-main`

valegagge:
Hi @Nicogene and @S-Dafarra ,
I'm sorry for not getting back to you sooner. 

@S-Dafarra : when we decided to add this functionality we knew we were adding a dependency to `icub-main`, but agreed with @Nicogene we decided to go ahead.
As you point out maybe it is too strong to make this dependency permanent. 
If I can post my few cents, I think we can keep the dependency from `Telemetrydumper` optional as @Nicogene suggested in the previous comment.

What do you think?

valegagge:
Hi @MSECode ,
this is the [commit](https://github.com/robotology/robometry/pull/190/commits/ed30a2cc57762bc796149b3a11126a8669755c8f) of the fis I mentioned [here](https://github.com/robotology/robometry/pull/190#issuecomment-2394702705).

MSECode:
> Hi @MSECode , this is the [commit](https://github.com/robotology/robometry/pull/190/commits/ed30a2cc57762bc796149b3a11126a8669755c8f) of the fis I mentioned [here](https://github.com/robotology/robometry/pull/190#issuecomment-2394702705).

Yes, I saw it and squashed to 1 single commit for the PR. 
Let's now decide how to managed the CMakeLists.txt file for what said in the thread and eventually add the fixes and proceed with the PR. Is it OK?

valegagge:
As discussed with @MSECode and @Nicogene , we'll organize a meeting with all the involved people to define if moving the device to `icub-main` or not or, in any case, point out the best solution.

S-Dafarra:
Let me add that the device is already dependent on ``yarp``, as it should be, so in any case, we already need for ``yarp`` to be compiled before. In general, I keep not liking to have ``icub-main`` as a dependency here because it hinders any usage of `robometry` within ``icub-main``. Still, I don't want to make a big fuzz out of this, so feel free to proceed with the solution you feel more suitable.

pattacini:
I didn't follow this discussion in the first place, but probably a better place for the I/F's would have been YARP itself. The implementation of the devices can stay in icub-main, but we usually put the I/F's in YARP right to address this kind of situation.

At any rate, we can discuss this topic with the ad-hoc meeting you scheduled.

traversaro:
> I didn't follow this discussion in the first place, but probably a better place for the I/F's would have been YARP itself. The implementation of the devices can stay in icub-main, but we usually put the I/F's in YARP right to address this kind of situation.
> 
> At any rate, we can discuss this topic with the ad-hoc meeting you scheduled.

We can discuss in person, but the idea was to have the interface somewhere maintained by iCub Tech to be able to quicky iterate if necessary.

src/telemetryDeviceDumper/CMakeLists.txt
```diff
@@ -20,12 +20,14 @@ if(ENABLE_telemetryDeviceDumper)
     add_definitions(-D_USE_MATH_DEFINES)
   endif()
 
+  find_package(iCubDev REQUIRED)
```
Nicogene, line 23: Have you bumped the patch version of icub-main? I would put here the required version that includes the new interface

src/telemetryDeviceDumper/TelemetryDeviceDumper.h
```diff
@@ -69,6 +74,7 @@ struct TelemetryDeviceDumperSettings {
  * | `logIAmplifierControl`     | bool     | -     | false     | No     | Enable the log of `motors_state::pwm` and `motors_state::currents` (http://yarp.it/git-master/classyarp_1_1dev_1_1IAmplifierControl.html).     |
  * | `logControlBoardQuantities` | bool     | -     | false     | No     | Enable the log of all the quantities that requires the attach to a control board (`logIEncoders`, `logITorqueControl`, `logIMotorEncoders`, `logIControlMode`, `logIInteractionMode`, `logIPidControl`, `logIAmplifierControl`). |
  * | `logILocalization2D` | bool     | -     | false     | No     | Enable the log of `odometry_data` (http://yarp.it/git-master/classyarp_1_1dev_1_1Nav2D_1_1ILocalization2D.html). |
+ * | `logIRawValuesPublisher` | bool |   -   | false    | No | Enable the log of raw values |
```
Nicogene, line 77: I would add a link to the interface stored in icub-main

src/telemetryDeviceDumper/TelemetryDeviceDumper.h
```diff
@@ -99,7 +105,8 @@ struct TelemetryDeviceDumperSettings {
  * | `PIDs::torque_error`       | [`yarp::dev::IPidControl::getPidErrors`](http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#aea29e0fdf34f819ac69a3b940556ba28) |
  * | `PIDs::torque_reference`   | [`yarp::dev::IPidControl::getPidReferences`](http://yarp.it/git-master/classyarp_1_1dev_1_1IPidControl.html#a29e8f684a15d859229a9ae2902f886da) |
  * | `PIDs::odometry_data   `   | [`yarp::dev::Nav2D::ILocalization2D::getEstimatedOdometry`](http://yarp.it/git-master/classyarp_1_1dev_1_1Nav2D_1_1ILocalization2D.html#a02bff57282777ce7511b671abd4c95f0) |
- *
+ * | `raw_data_values`          | [`iCub::debugLibrary::IRawValuesPublisher::getRawDataMap`]
```
Nicogene, line 108: Same as above

Nicogene (changes requested):
Some small changes requested

src/telemetryDeviceDumper/CMakeLists.txt
MSECode: Yes, actually as suggested by @pattacini we bumped the minor version.

Nicogene: Great! It should be then `find_package(iCubDev 2.7.0 REQUIRED)`, and I would also add it in the readme documentation of the device that that version of `icub-main` is required

src/telemetryDeviceDumper/TelemetryDeviceDumper.h
MSECode: done

src/telemetryDeviceDumper/TelemetryDeviceDumper.h
MSECode: done

src/telemetryDeviceDumper/CMakeLists.txt
MSECode: done

Nicogene (approved):

PR #194, opened 2025-03-31
Fix compilation under conda

Nicogene:
As per title


traversaro:
Probably this involves some more changes, but now that we do not support apt dependencies for Ubuntu 20.04 apt anymore, could it make sense that we just consume nlohmann_json as any other dependency instead of using FetchContent? The version available in Ubuntu 22.04 with apt dependencies is 3.10 (see https://repology.org/project/nlohmann-json/versions), that I think is new enough.

PR #196, opened 2025-04-10
Add motor temperature streaming in the `TelemetryDeviceDumper` 

MSECode:
This PR, which is strictly correlated to the modifications done by @GiulioRomualdi in https://github.com/robotology/yarp/pull/3188 and to the information added in [this issue](https://github.com/icub-tech-iit/five/issues/314), allow to add to the `motor_states` in the `MATLAB` files given by the `TelemetryDeviceDumper` the motor temperatures.
As added with a note in the `README.md`, this is still an unstable feature that for working needs manual changes to the compilation of icub-main.
Thus, I suggest to use it with care and mainly by developers or under their supervision for the initial compilation, since it might lead to compilation issues with icub-main. 

cc: @valegagge

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,7 +480,11 @@ The configuration can be saved **to a json file**
 
 ## TelemetryDeviceDumper
 
-The `telemetryDeviceDumper` is a [yarp device](http://yarp.it/git-master/note_devices.html) that has to be launched through the [`yarprobotinterface`](http://yarp.it/git-master/yarprobotinterface.html) for dumping quantities from your robot(e.g. encoders, velocities etc.) in base of what specified in the configuration. It currently needs icub-main version equal or higher than `2.7.0` Specificially this is needed when enabling the parameter `logIRawValuesPublisher`, which is used for dumping any type of raw data values coming from the low level, e.g. raw encoder data. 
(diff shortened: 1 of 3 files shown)


Nicogene (commented):
Thanks for the PR @MSECode, the CI is happy so the dependency is not build-time but run-time?

valegagge:
Hi @MSECode and @Nicogene ,
What do you think about putting this PR in Draft while we are waiting for the YArp PR to be merged?

traversaro:
> Thanks for the PR @MSECode, the CI is happy so the dependency is not build-time but run-time?

Yes, the dependency is on the runtime **behaviour** of reading temperature. With the existing YARP, calling getTemperature at high frequency is not possible as it will trigger a lot of **blocking** RPC alls, that will slowdown the logger and flood the network. However, this is not a problem as long as the `logIMotorTemperatures` parameter is not used, so I guess there is no harm in merging this PR even if the corresponding YARP PR is not merged?

valegagge:
> > Thanks for the PR @MSECode, the CI is happy so the dependency is not build-time but run-time?
> 
> Yes, the dependency is on the runtime **behaviour** of reading temperature. With the existing YARP, calling getTemperature at high frequency is not possible as it will trigger a lot of **blocking** RPC alls, that will slowdown the logger and flood the network. However, this is not a problem as long as the `logIMotorTemperatures` parameter is not used, so I guess there is no harm in merging this PR even if the corresponding YARP PR is not merged?

Yep!!!

Nicogene (approved):

PR #203, opened 2026-07-21
telemetryDeviceDumper: add compatibility with YARP 4

Nicogene:
It keeps backward compatibility with YARP 3. @valegagge I think we can merge right away this and drop the development on yarp4 branch.


valegagge (approved):
Good idea!!

